PR #14, opened 2019-03-05
Makefile: find preCICE using pkg-config

MakisH:
This is necessary for moving preCICE from SCons to CMake. Since the C bindings are expected in a different location than before, we decided to only support the new location, for simplicity.

This breaks the out-of-the-box compatibility with SCons and the user is referred to this commit in case she
wants to keep using preCICE build with SCons.

This PR also fixes indentation and mixed tabs/spaces issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,46 +1,47 @@
-# Specify the locations of: the original CCX source, SPOOLES and ARPACK
+# See our wiki for getting the CalculiX dependencies:
+# https://github.com/precice/calculix-adapter/wiki/Installation-instructions-for-CalculiX
+# Set the following variables before building:
+# Path to original CalculiX source (e.g. $(HOME)/ccx_2.15/src )
 CCX             = $(HOME)/PathTo/CalculiX/ccx_2.15/src
+# Path to SPOOLES main directory (e.g. $(HOME)/SPOOLES.2.2 )
 SPOOLES         = $(HOME)/PathTo/SPOOLES
+# Path to ARPACK main directory (e.g. $(HOME)/ARPACK )
 ARPACK          = $(HOME)/PathTo/ARPACK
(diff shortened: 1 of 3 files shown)


MakisH:
Reviewers: please try if this works with your setups and document here what we need to update.
* @shkodm: could you also port these changes to the systemtests?
* @KyleDavisSA: do we need to update any other scripts/documentation/tutorials?

MakisH:
Please note that, if you are using preCICE as a static library, you should wait for [this issue](https://github.com/precice/precice/issues/313) to be resolved.

MakisH:
Thanks for the comments!

> The changes work (and the wiki documentation probably updated to understand them easier) if I made sure:
> 
>     1. preCICE v1.4 was installed in a seperate directory from the "/usr/local" directory. This was set using -DCMAKE_INSTALL_PREFIX=/some/user/specified/path. Should this be the norm? There was no problems with running "make install" after that.

Note that the (default) `/usr/local` is a system directory, which requires root permissions to be written (e.g. `sudo make install`). In principle, the user can choose any path, and that's why we now have the prefix information in the `libprecice.pc` and the respective CMake config files.

>     2. Any problems with finding files were solved with "export PKG_CONFIG_PATH=/path/to/pkgconfig" where the file "libprecice.pc" is located.

That's correct and intended.

>     3. My makefile was complaining about seperators. I had to delete the spaces and put tabs. I am not sure if this problem would be repeated?

Yes, this is indeed a problem, as Make (at least older versions) only accept tabs. I will fix it.

MakisH:
The Travis succeeds incorrectly for this PR (as described in https://github.com/precice/systemtests/issues/50) and it should fail after we merge this. I will update the system tests respectively.

PR #16, opened 2019-06-19
memory leak and missing allocation

nkr1729:


nkr1729:
the 2nd commit makes changes to take into account the dimension in which the coupling is happening.

adapter/PreciceInterface.c
```diff
@@ -394,15 +396,15 @@ void PreciceInterface_ConfigureNodesMesh( PreciceInterface * interface, Simulati
...
-	getNodeCoordinates( interface->nodeIDs, interface->numNodes, sim->co, sim->vold, sim->mt, interface->nodeCoordinates );
+	interface->nodeCoordinates = malloc( interface->numNodes * interface->dim * sizeof( double ) );
+	getNodeCoordinates( interface->nodeIDs, interface->numNodes, interface->dim, sim->co, sim->vold, sim->mt, interface->nodeCoordinates );
 
 	if( interface->nodesMeshName != NULL )
 	{
 		//printf("nodesMeshName is not null \n");
 		interface->nodesMeshID = precicec_getMeshID( interface->nodesMeshName );
-		//interface->preciceNodeIDs = malloc( interface->numNodes * sizeof( int ) );
-		interface->preciceNodeIDs = malloc( interface->numNodes * 3 * sizeof( int ) );
```
nkr1729: why was `preciceNodeIDs` x3?

KyleDavisSA:
Do you have a working example with taking the dimension into account. The cylinderFlap test case fails when using 2D.
Screenshot from 2019-07-09 10-06-13

nkr1729:
That is because openfoam is not 2D. The mesh is 3D. To test the 2D interface you need actual 2D meshes on both sides. the screenshot below is calculix with [fluidity](https://github.com/fluidityproject). I use shell elements (CPE3) in calculix and triangular elements in fluidity.

flap

The publicly available version of fluidity doesn't have preCICE integration. So I cannot share my example with you. However, you can write a small python script that simulates the CSM test cases (cantilever loading) from the turek-hron paper. Or couple with a pure 2D fluid solver that has preCICE integration.

KyleDavisSA:
I see, fluidity uses a purely 2D mesh and not some quasi-3D mesh like openFOAM. I get correct results with them in 3D, so it is working properly in that case.

nonlingeo_precice.c
```diff
@@ -2632,6 +2632,7 @@ void nonlingeo_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **l
     	NNEW(fn,double,mt**nk);
     	NNEW(stn,double,6**nk);
 	    NNEW(stx,double,6*mi[0]**ne);
+    	NNEW(inum,ITG,*nk);
```
MakisH: What is this and why do we need it?

adapter/PreciceInterface.c
MakisH: This originates from the nearest-projection mapping additions. Your change looks correct to my understanding and I would even remove the commented-out part if @KyleDavisSA agrees.

nkr1729:
both `stn` and `inum` are calculix variables unrelated to preCICE. from the calculix manual,

stn

inum

I'm not sure why this wasn't an issue before, but `inum` needs to be allocated before calling `results` (from trial and comparing with other parts of `nonlingeo_precice.c`).

allocation of `stn` was already there and I can't remember whether it was necessary or not. I think I just added the free statement based on valgrind's output.

PR #19, opened 2019-07-15
read/write multiple fields via an interface

### LINKED ISSUE #17: make readData and writeData array
We are iterating over numReadData.

https://github.com/precice/calculix-adapter/blob/10938d846aa99e1b5a7c8a38dbf4cf5082c586aa/adapter/PreciceInterface.c#L467

But, readData is not an array.

https://github.com/precice/calculix-adapter/blob/10938d846aa99e1b5a7c8a38dbf4cf5082c586aa/adapter/PreciceInterface.h#L71

Same goes for writeData. If we have a config like this,

```
participants:

    Calculix:
        interfaces:
        - nodes-mesh: Calculix_Mesh
          patch: interface
          read-data: [Read0,AnotherRead0]
          write-data: [Write0,AnotherWrite0]

precice-config-file: ../precice-config.xml
```

`numRead/WriteData` will be 2.

### LINKED ISSUE #18: Remove appending index from readData and writeData labels
readData and writeData names are stored inside `PreciceInterface`.

https://github.com/precice/calculix-adapter/blob/10938d846aa99e1b5a7c8a38dbf4cf5082c586aa/adapter/PreciceInterface.h#L71

We have an array of interfaces to isolate these from getting mixed between interfaces.

https://github.com/precice/calculix-adapter/blob/10938d846aa99e1b5a7c8a38dbf4cf5082c586aa/adapter/PreciceInterface.c#L33

So from CalculiX itself, we don't need this index.
```
participants:

    ccx1:
        interfaces:
        - nodes-mesh: interface0
          patch: patch0
          read-data: [Read,AnotherRead]
          write-data: [Write,AnotherWrite]
        - nodes-mesh: interface1
          patch: patch1
          read-data: [Read,AnotherRead]
          write-data: [Write,AnotherWrite]

precice-config-file: ../precice-config.xml
```

And on the precice config side,
```
    <data:vector name="Read"  />
    <data:vector name="AnotherRead"  />
    <data:vector name="Write"  />
    <data:vector name="AnotherWrite"  />

    <mesh name="interface0">
      <use-data name="Read" />
      <use-data name="AnotherRead" />
      <use-data name="Write" />
      <use-data name="AnotherWrite" />
    </mesh>

    <mesh name="interface1">
      <use-data name="Read" />
      <use-data name="AnotherRead" />
      <use-data name="Write" />
      <use-data name="AnotherWrite" />
    </mesh>
```

Did I miss a case where we need the index to make things work?

nkr1729:
a detailed description of things this PR fixes can be found in the issues

fixes #17 
fixes #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,8 +122,8 @@ participants:
         interfaces:
         - nodes-mesh: Calculix_Mesh
           patch: interface
-          read-data: [Forces0]
-          write-data: [DisplacementDeltas0]
+          read-data: [Forces]
+          write-data: [DisplacementDeltas]
 
 precice-config-file: ../precice-config.xml
 ```
@@ -132,7 +132,7 @@ The adapter allows to use several participants in one simulation (e.g. several i
 One participant may have several FSI interfaces. Note that each interface specification starts with a dash.  
 For FSI simulations the mesh type of an interface is always "nodes-mesh", i.e. the mesh is defined node-wise, not element-wise. The name of this mesh, "Calculix_Mesh", must match the mesh name given in the preCICE configuration file.  
 For defining which nodes of the CalculiX domain belong to the FSI interface, a node set needs to be defined in the CalculiX input files. The name of this node set must match the name of the patch (here: "interface").  
-In the current FSI example, the adapter reads forces from preCICE and feeds displacement deltas (not absolute displacements, but the change of the displacements relative to the last time step) to preCICE. This is defined with the keywords "read-data" and "write-data", respectively. The names (here: "Forces0" and "DisplacementDeltas0") again need to match the specifications in the preCICE configuration file. Note that the "0" appending the data names is not optional. Since it is possible to define several interfaces, the data specifications need to be numbered starting from zero (e.g. Forces0, Forces1, Forces2, ...). In the current example, the coupled fluid solver expects displacement deltas instead of displacements. However, the adapter is capable of writing either type. Just use "write-data: [Displacements0]" for absolute displacements rather than relative changes being transferred in each time step. Note that the square brackets imply that several read- and write-data types can be used on a single interface. This is not needed for FSI simulations (but for CHT simulations). Lastly, the "precice-config-file" needs to be identified including its location. In this example, the file is called *precice-config.xml* and is located one directory above the folder, in which the YAML configuration file lies.
+In the current FSI example, the adapter reads forces from preCICE and feeds displacement deltas (not absolute displacements, but the change of the displacements relative to the last time step) to preCICE. This is defined with the keywords "read-data" and "write-data", respectively. The names (here: "Forces" and "DisplacementDeltas") again need to match the specifications in the preCICE configuration file. In the current example, the coupled fluid solver expects displacement deltas instead of displacements. However, the adapter is capable of writing either type. Just use "write-data: [Displacements]" for absolute displacements rather than relative changes being transferred in each time step. Note that the square brackets imply that several read- and write-data types can be used on a single interface. This is not needed for FSI simulations (but for CHT simulations). Lastly, the "precice-config-file" needs to be identified including its location. In this example, the file is called *precice-config.xml* and is located one directory above the folder, in which the YAML configuration file lies.
 
 ### CalculiX Case Input File
 An exemplary CalculiX case input file may look like the following:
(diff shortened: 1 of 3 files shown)


KyleDavisSA:
I have been testing the PR with the various tutorials available with CalculiX. Regarding issue #17, I don't remember multiple readData and writeData arrays being a problem. 

Regarding #18 , I am testing the CHT heat exchanger case where both the inner and outer fluid domains have the same name, and the only difference is the interface name they belong to. Unless I am misunderstanding the issue? Could you provide an example that you tested the changes with?

nkr1729:
Looking at the CHT case, that works because `Sink-Temperature-` label is stored in `readData` https://github.com/precice/calculix-adapter/blob/10938d846aa99e1b5a7c8a38dbf4cf5082c586aa/adapter/PreciceInterface.c#L492
and, `Heat-Transfer-Coefficient` is stored in `kDeltaReadDataID`
https://github.com/precice/calculix-adapter/blob/10938d846aa99e1b5a7c8a38dbf4cf5082c586aa/adapter/PreciceInterface.c#L500
So that bypasses #17. I hadn't noticed this before and I'm not sure why a new variable was created to store this label. Also, you can see that only `Forces`, `Displacements`, and `DisplacementDeltas` require the index at the end.
https://github.com/precice/calculix-adapter/blob/10938d846aa99e1b5a7c8a38dbf4cf5082c586aa/adapter/PreciceInterface.c#L503
And that bypasses #18. Again, this is not very clear as in the Readme it doesn't say that only these 2 variables require an index appended.

So, I think you do understand the issues. But, the issues don't show up in the CHT example. If you create a set of read/write fields with anything other than `Sink-Temperature-` and `Heat-Transfer-Coefficient`, both the issues will show up.

I've the same problem sharing an example, because I use fluidity. Perhaps you can take the cylinderflap case and add `DisplacementDeltas0` to the array of `write-data` fields.
https://github.com/precice/tutorials/blob/master/FSI/cylinderFlap/OpenFOAM-CalculiX/config.yml
Please add it before `Displacements0` and the simulation should fail. I haven't tried. But my guess is, the body won't move.
1) In the output calculix prints, you should see that only the first field, `DisplacementDeltas0` is found.
2) OpenFOAM won't get anything from `Displaments0` because calculix will only write `DisplacementDeltas0`, the first field in the list.
Whereas, with the PR,
1) calculix will print in the output that it detected both the fields.
2) And write both `Displaments` and `DisplacementDeltas` to the interface. OpenFOAM can use what it needs.
Notice that in the PR you don't need indexes at the end. You'll also have to make appropriate changes in the precice-config file.

KyleDavisSA:
I performed a test where I added `write-data:[Displacements,DisplacementDeltas]` to the config.yml files. I also added the `Displacements` to the `precice-config.xml` file as a data vector. I assume that having multiple of these fields is what is needed. Physically, there is no scenario that needs this, however that does solve the issue if temperatures are exchanged on a moving surface.

The config.yml looks like:
```
        - nodes-mesh-with-connectivity: Calculix_Mesh
          patch: interface
          read-data: Forces
          write-data: [Displacements,DisplacementDeltas]
```

and parts of the precice-config.xml now has:
```
      <data:vector name="Forces"/>
      <data:vector name="DisplacementDeltas"/>
      <data:vector name="Displacements"/>
```
```
      <mesh name="Calculix_Mesh">
         <use-data name="DisplacementDeltas"/>
         <use-data name="Displacements"/>
         <use-data name="Forces"/>
      </mesh>
```
```
      <participant name="Calculix">
         <use-mesh name="Calculix_Mesh" provide="yes"/>
        <use-mesh name="Fluid-Mesh-Faces" from="Fluid"/>
         <write-data name="DisplacementDeltas" mesh="Calculix_Mesh"/>
         <write-data name="Displacements" mesh="Calculix_Mesh"/>
         <read-data  name="Forces" mesh="Calculix_Mesh"/>
      </participant>
```

This is not an exhaustive list of changes, but this setup is what you are refering to? These changes work and the simulation runs with the PR.

nkr1729:
There is a small mistake compared to what I wrote. Instead of `write-data: [Displacements,DisplacementDeltas]` change it to `write-data: [DisplacementDeltas,Displacements]`. Calculix takes only the 1st value. So we want what to push what OpenFOAM expects to the second position.

Yes, having multiple fields is one of the purposes of this PR. Sure, OpenFOAM's version of FSI doesn't need this. But I think there might be cases/users who want to. I use this because instead of computing `Velocity` on the fluid side from `Displacements`, I transfer that field also from CalculiX.

PR #23, opened 2019-11-27
save state after every coupling iteration in explicit coupling

### LINKED ISSUE #8: DisplacementDeltas not implemented for explicit coupling?
https://github.com/precice/calculix-adapter/blob/master/adapter/PreciceInterface.c

It seems like `sim->coupling_init_v` is only updated for implicit coupling (through `Precice_WriteIterationCheckpoint`), but never if only explicit coupling is used. 

@arusch Are we overseeing something? 

Thanks!

### LINKED ISSUE #18: Remove appending index from readData and writeData labels

nkr0:
fix #8 

A small hack to make DisplacementDeltas work for explicit coupling as well. My question however is, is DisplacementDeltas change in displacement within a sub-cycle or change in displacement relative to the displacement at the beginning of the coupling step. I hope I'm clear here. The changes here assume the former.

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -48,7 +48,7 @@ void Precice_Setup( char * configFilename, char * participantName, SimulationDat
 	Precice_InitializeData( sim );
 
 	// find if coupling is implicit or explicit. Implicit coupling will return true at the very beginning and explicit will always return false
-	sim->coupling_explicit = !Precice_IsWriteCheckpointRequired();
+	sim->coupling_implicit = Precice_IsWriteCheckpointRequired();
 
 }
 
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -114,6 +114,11 @@ bool Precice_IsWriteCheckpointRequired()
 	return precicec_isActionRequired( "write-iteration-checkpoint" );
 }
 
+bool Precice_IsCouplingTimestepComplete()
+{
+	return precicec_isCouplingTimestepComplete();
+}
+
 void Precice_FulfilledReadCheckpoint()
 {
 	precicec_fulfilledAction( "read-iteration-checkpoint" );
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -47,6 +47,9 @@ void Precice_Setup( char * configFilename, char * participantName, SimulationDat
 	// Initialize coupling data
 	Precice_InitializeData( sim );
 
+	// find if coupling is implicit or explicit. Implicit coupling will return true at the very beginning and explicit will always return false
+	sim->coupling_explicit = !Precice_IsWriteCheckpointRequired();
+
 }
 
 void Precice_InitializeData( SimulationData * sim )
--- a/nonlingeo_precice.c
+++ b/nonlingeo_precice.c
@@ -1114,10 +1114,10 @@ void nonlingeo_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **l
 	  
 	  memcpy(&vini[0],&vold[0],sizeof(double)*mt**nk);
 
-	  if( Precice_IsWriteCheckpointRequired() )
+	  if( Precice_IsWriteCheckpointRequired() || simulationData.coupling_explicit)
       	  {
           	Precice_WriteIterationCheckpoint( &simulationData, vini );
-          	Precice_FulfilledWriteCheckpoint();
+          	if( Precice_IsWriteCheckpointRequired() ) Precice_FulfilledWriteCheckpoint();
           }
 	  
 	  for(k=0;k<*nboun;++k){xbounini[k]=xbounact[k];}
--- a/nonlingeo_precice.c
+++ b/nonlingeo_precice.c
@@ -1114,7 +1114,7 @@ void nonlingeo_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **l
 	  
 	  memcpy(&vini[0],&vold[0],sizeof(double)*mt**nk);
 
-	  if( Precice_IsWriteCheckpointRequired() || simulationData.coupling_explicit)
+	  if( Precice_IsWriteCheckpointRequired() == simulationData.coupling_implicit)
       	  {
           	Precice_WriteIterationCheckpoint( &simulationData, vini );
           	if( Precice_IsWriteCheckpointRequired() ) Precice_FulfilledWriteCheckpoint();
--- a/nonlingeo_precice.c
+++ b/nonlingeo_precice.c
@@ -1089,6 +1089,8 @@ void nonlingeo_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **l
   /* Adapter: Create the interfaces and initialize the coupling */
   printf("About to enter preCICE setup in Calculix with names %s and %s \n", preciceParticipantName, configFilename);
   Precice_Setup( configFilename, preciceParticipantName, &simulationData );
+  // Initialize for the first step because Precice_IsCouplingTimestepComplete will return false until the end of the first step
+  if( !simulationData.coupling_implicit ) Precice_WriteIterationCheckpoint( &simulationData, vini );
   
   /* Adapter: Give preCICE the control of the time stepping */
   while( Precice_IsCouplingOngoing() ){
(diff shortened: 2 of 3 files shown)


MakisH:
I am not sure I understand the solution. Why store a checkpoint in explicit coupling? Maybe you can extract only the relevant bit and call it always, independent of the checkpointing?

nonlingeo_precice.c
```diff
@@ -1114,10 +1114,10 @@ void nonlingeo_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **l
 	  
 	  memcpy(&vini[0],&vold[0],sizeof(double)*mt**nk);
 
-	  if( Precice_IsWriteCheckpointRequired() )
+	  if( Precice_IsWriteCheckpointRequired() || simulationData.coupling_explicit)
```
MakisH: This would then almost always be true, right? I understnad that you always want checkpoints to be stored (to trigger some internal update).

adapter/PreciceInterface.c
```diff
@@ -47,6 +47,9 @@ void Precice_Setup( char * configFilename, char * participantName, SimulationDat
 	// Initialize coupling data
 	Precice_InitializeData( sim );
 
+	// find if coupling is implicit or explicit. Implicit coupling will return true at the very beginning and explicit will always return false
+	sim->coupling_explicit = !Precice_IsWriteCheckpointRequired();
```
MakisH: If you need such a boolean, it would be better to make is `coupling_implicit`. Usually we assume that implicit coupling is the special case that requires special treatment. You could then still write `if (!coupling_implicit)`.

But I think you don't need this check in the first place, it should be enough to call what you want always.

MakisH (changes requested):
Thanks for the contributions! :D I hope these comments help express a bit better in the fix what I think you actually want to do.

nonlingeo_precice.c
nkr0: For explicit coupling, yes.

adapter/PreciceInterface.c
nkr0: sure, `!coupling_implicit` or `coupling_explicit` is no problem. But a check is needed. Because the call I want to happen always is only in the case of explicit coupling.

nkr0:
> I am not sure I understand the solution. Why store a checkpoint in explicit coupling? Maybe you can extract only the relevant bit and call it always, independent of the checkpointing?

There is more information in issue #8. It's not exactly a "checkpoint". We just want the `memcpy` line from the write function,
https://github.com/precice/calculix-adapter/blob/b01641e40c11e067f66cc41fbcb0253d08290ff0/adapter/PreciceInterface.c#L140-L154
so that the reference point w.r.t. which we compute DisplacementDeltas, `sim->coupling_init_v` (`v_init`) gets updated in every iteration for only explicit cases.
https://github.com/precice/calculix-adapter/blob/b01641e40c11e067f66cc41fbcb0253d08290ff0/adapter/CCXHelpers.c#L135
So yes, we could extract the `memcpy` line and write it. But the only differences between calling `writecheckpoint` and calling `memcpy` are the 2 lines of saving `theta` and `dtheta`. It really doesn't make any difference.

But, all of this is only if my understanding of the definition of DisplacementDeltas is correct. What is the reference point for DisplacementDeltas? Is it the Displacement from the last sub-cycle? Like it is implemented now. Or, is it the Displacement from the beginning of the coupling iteration? Like how it is for implicit cases, where the reference point is set at the beginning of a new coupling iteration, controlled by the `IsWriteCheckpointRequired` call. Explicit coupling cases also synchronise at regular coupling intervals right? Either fixed or controlled by one of the participants.

KyleDavisSA:
A question from #8 

> Here the previous solution is added to coordinates and the comment says it is for restarts. Although I'm not really sure how correct that is. Displacement is the solution from a loaded condition. But changing the coordinates really alters the solid. It can stay in the displaced state without any load acting on it.

That is correct, but I guess the description of a restart is vague. The "restart" is when the simulation must be stopped and restarted from the same point in time. An example would be if the fluid mesh is too distorted. The simulation can be stopped, the fluid mesh re-meshed, and the simulation restarted from the last time step. This then does involve restarting CalculiX from the previous loaded condition.

> so that the reference point w.r.t. which we compute DisplacementDeltas, sim->coupling_init_v (v_init) gets updated in every iteration for only explicit cases.

Before `void Precice_WriteIterationCheckpoint( SimulationData * sim, double * v )`, the variable `double * v ` is updated with the displacement at the end of the previous iteration `vold`

https://github.com/precice/calculix-adapter/blob/b01641e40c11e067f66cc41fbcb0253d08290ff0/nonlingeo_precice.c#L1115-L1121

This is just to clarify that DisplacementDeltas should be the displacement at the current time - displacement at the end of the previous timestep.

nkr0:
The part about restart from #8 is only something I misunderstood back then. Nothing to do with this PR or issue #8 itself.
> This is just to clarify that DisplacementDeltas should be the displacement at the current time - displacement at the end of the previous timestep.

That is correct, and that will be the case once the PR is merged. First we copy `vold` into `vini`. And then `vini` is saved as displacement at the previous time step.
https://github.com/precice/calculix-adapter/blob/0e035eaf4e702898143111fa4d9c58a155a37b38/nonlingeo_precice.c#L1115-L1119
After the solution for the current time step is computed, `v` is copied into `vold` and we call `WriteCouplingData`.
https://github.com/precice/calculix-adapter/blob/0e035eaf4e702898143111fa4d9c58a155a37b38/nonlingeo_precice.c#L2664-L2666
Inside `WriteCouplingData` we compute `DisplacementDelta`,
https://github.com/precice/calculix-adapter/blob/0e035eaf4e702898143111fa4d9c58a155a37b38/adapter/PreciceInterface.c#L301-L303
where `v_init` is the old solution we stored at the beginning of the iteration and `v` is a direct pointer access to the current solution CalculiX had just finished calculating.
https://github.com/precice/calculix-adapter/blob/0e035eaf4e702898143111fa4d9c58a155a37b38/adapter/CCXHelpers.c#L125-L137

nkr0:
Also, I've realised that for explicit coupling since there are no checkpoints, the concerns I raised above are also invalid. So except for the part where @MakisH said `coupling_implicit` is a better boolean choice than `coupling_explicit`, I think this PR is correct.

uekerman:
I am sorry for the very late reply here.

I appreciate the contribution a lot, but I am not 100% happy with the current solution. It was already a bit hacky before (current master) and now it seems to get too complicated.

Up front, to clarify again: `DisplacementDeltas` should be relative to the displacements from the last coupling timestep / coupling window. This definition is (and needs to be) independent of whether explicit or implicit coupling is used. 

Explicitly storing whether a implicit or explicit coupling scheme is used (through a variable such as `coupling_explicit`) is something that is not intended by the preCICE API. Whenever you feel you need to do this, there has to be a cleaner way (or the design assumptions of the preCICE API are wrong). I agree that fully understanding this is an hard entry barrier and thus, a weak spot of preCICE.

I guess the cleaner solution here is to use two separate variables:
* One to checkpoint the solution, needs to be displacements of the full domain (or maybe also velocity values etc depending on the time stepping method used)
  * read and write this variable in `write/readIterationCheckpoint`
* One to store the referential displacement values for `DisplacementDeltas`. This could be just stored in the "preCICE data structure", so only values at the coupling interface.
  * write this variable in `isTimestepComplete`, use it when writing `DisplacementDeltas` to preCICE.
 
Understandable?

nkr0:
Yes, understood. However, check the bit below,
https://github.com/precice/calculix-adapter/blob/d1a4f5aa20eb39e9021b2bebdd0c7a5be485991c/nonlingeo_precice.c#L1119-L1125
in implicit coupling, when does `isTimestepComplete` return `true`? Is it **only** at the end of a **converged** coupling time-step, or every time the coupling iteration reaches the end of the coupling time-step (even when `readIterationCheckpoint` is going to restart the iteration)? If it is the former, yeah we can use `isTimestepComplete` for `DisplacementDeltas`, and `isWriteCheckpointRequired` for everything else. But, in the latter case we cannot and will need an explicit/implicit switch to in addition to `isTimestepComplete`. This doubt lead me to create separate cases for implicit and explicit.

uekerman:
> in implicit coupling, when does `isTimestepComplete` return `true`? Is it **only** at the end of a **converged** coupling time-step, or every time the coupling iteration reaches the end of the coupling time-step (even when `readIterationCheckpoint` is going to restart the iteration)?

It only returns true if the timestep is completely completed, so converged (for implicit coupling) and last sub timestep. So the former.

If it is really the latter it would be a bug in preCICE.

PR #40, opened 2020-06-25
Migrate to CMake

fsimonis:
This PR migrates the build system of the CCX adapter to CMake, which handles the complete build of the adapter. The result is a statically linked executable.

To build:
1. Install preCICE
2. Install LAPACK (dependency of ARPACK)
3. `cmake .`
4. `make`

To update:
1. Update `CCX_VERSION` in `CMakeLists.txt`
2. Update patched ~~`CalculiX.h` and~~ `ccx_X.YY.c`
3. Update adapter files

This should also allow us to create a relatively straight forward CI pipeline. We just need an image with precice installed.


fsimonis:
As it turns out, both arpack and spooles are available through various package managers, thus this PR can be simplified.
Closing until we have a plan how to proceed with this.

PR #41, opened 2020-07-11
Extend the Makefile to support dependencies from APT by default

MakisH:
The current default does not work without modifications in any scenario: The user is first asked to edit the Makefile to specify the paths to SPOOLES, ARPACK, and yaml-cpp.

This PR makes it easier to build the adapter in a Debian-based system, after simply getting SPOOLES, ARPACK, and yaml-cpp from APT. For users that have these dependencies elsewhere, they still need to set the dependencies themselves.

Furthermore, it sets the suggested location for CCX and the dependencies to `$(HOME)`, a pattern often observed among new users.

This also splits each dependency to an `*_INCLUDE` and an `*_LIBS` part, making it more flexible.

After merging:
- [ ] Update the `Makefile variables` section in [Building the Adapter](https://github.com/precice/calculix-adapter/wiki/Building-the-Adapter)
- [x] Update the system tests @Eder-K 
   - Update the Makefile
   - Replace building SPOOLES/ARPACK/yaml-cpp from source with getting them from APT

@fsimonis: Please check if my changes make sense (there is a high chance that I have left some typo here, even though I tested both ways in two systems).

@KyleDavisSA: Try building with your favourite dependency paths, maybe also following the updated [building instructions](https://github.com/precice/calculix-adapter/wiki/Installation-instructions-for-CalculiX).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,20 +5,20 @@
 CCX             = $(HOME)/CalculiX/ccx_2.16/src
 
 ### Change these if you built SPOOLES, ARPACK, or yaml-cpp from source ###
-# SPOOLES include directory (e.g. $(HOME)/SPOOLES.2.2 )
-SPOOLES_INCLUDE   = /usr/include/spooles/
+# SPOOLES include flags (e.g. -I$(HOME)/SPOOLES.2.2 )
+SPOOLES_INCLUDE   = -I/usr/include/spooles/
 # SPOOLES library flags (e.g. $(HOME)/SPOOLES.2.2/spooles.a)
-SPOOLES_LIBS      = -L/usr/lib/x86_64-linux-gnu/ -lspooles
+SPOOLES_LIBS      = -lspooles
 #
-# ARPACK include directory (e.g. $(HOME)/ARPACK)
-# ARPACK_INCLUDE  =
+# ARPACK include flags (e.g. -I$(HOME)/ARPACK)
+ARPACK_INCLUDE    =
 # ARPACK library flags (e.g. $(HOME)/ARPACK/libarpack_INTEL.a)
-ARPACK_LIBS       = -L/usr/lib/ -larpack -llapack -lblas
+ARPACK_LIBS       = -larpack -llapack -lblas
 #
-# yaml-cpp include path (e.g. $(HOME)/yaml-cpp/include)
-YAML_INCLUDE      = /usr/include/
+# yaml-cpp include flags (e.g. -I$(HOME)/yaml-cpp/include)
+YAML_INCLUDE      = -I/usr/include/
 # yaml-cpp library flags (e.g. -L$(HOME)/yaml-cpp/build -lyaml-cpp)
-YAML_LIBS         = -L/usr/lib/x86_64-linux-gnu/ -lyaml-cpp
+YAML_LIBS         = -lyaml-cpp
 
 # Get the CFLAGS and LIBS from pkg-config (preCICE version >= 1.4.0).
 # If pkg-config cannot find the libprecice.pc meta-file, you may need to set the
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,13 +2,23 @@
...
+#
+# ARPACK include directory (e.g. $(HOME)/ARPACK)
+# ARPACK_INCLUDE  =
+# ARPACK library flags (e.g. $(HOME)/ARPACK/libarpack_INTEL.a)
+ARPACK_LIBS       = -L/usr/lib/ -larpack -llapack -lblas
+#
+# yaml-cpp include path (e.g. $(HOME)/yaml-cpp/include)
+YAML_INCLUDE      = /usr/include/
+# yaml-cpp library flags (e.g. -L$(HOME)/yaml-cpp/build -lyaml-cpp)
+YAML_LIBS         = -L/usr/lib/x86_64-linux-gnu/ -lyaml-cpp
```
fsimonis, line 21: Are the `-L` flags necessary here? If the linker cannot find these libraries out of the box then there will be runtime errors as their locations are unknown. (If they are shared libraries)

Makefile
```diff
@@ -2,13 +2,23 @@
...
 # Set the following variables before building:
 # Path to original CalculiX source (e.g. $(HOME)/ccx_2.16/src )
-CCX             = $(HOME)/PathTo/CalculiX/ccx_2.16/src
-# Path to SPOOLES main directory (e.g. $(HOME)/SPOOLES.2.2 )
-SPOOLES         = $(HOME)/PathTo/SPOOLES
-# Path to ARPACK main directory (e.g. $(HOME)/ARPACK )
-ARPACK          = $(HOME)/PathTo/ARPACK
-# Path to yaml-cpp prefix (e.g. $(HOME)/yaml-cpp, should contain "include" and "build")
-YAML            = $(HOME)/PathTo/yaml-cpp
+CCX             = $(HOME)/CalculiX/ccx_2.16/src
```
fsimonis, line 5: We could deduce this from the location of the Makefile as described [here](https://stackoverflow.com/a/18137056/11633170). That would be one less thing to worry about

_untested suggestion:_
```suggestion
CCX             = $(strip $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
```

Makefile
MakisH: They are not necessary here, but I am still providing them explicitly as an example for the user. But indeed, removing them would be more robust. I will remove them.

Makefile
MakisH: If I understand correctly, this will give the current directory (i.e. `calculix-adapter`). What we want is the path to the vanilla CalculiX directory, which can be anywhere in the system.

fsimonis: Right, sorry! 

Maybe it would be simpler to specify where to extract the sources instead. There are only a few MB and would remove another customization point in the installation procedure.
We could even provide a `download_ccx.sh` for that purpose.

MakisH: Good point, I made this clearer in the [documentation](https://github.com/precice/calculix-adapter/wiki/Installation-instructions-for-CalculiX#get-the-source).

However, I am not sure if providing a `download_ccx.sh` script would be a good idea: it would probably confuse some users (especially those that have CalculiX already), or at least more than the current variant confuses most users.

MakisH:
@fsimonis thanks for the useful feedback. I removed the `-L`, moved the `-I` into the user-defined part (consistency + avoid dummy `-I.`), and added a hint for where to get `CCX` in the documentation.

Makefile
```diff
@@ -2,13 +2,23 @@
...
+
+### Change these if you built SPOOLES, ARPACK, or yaml-cpp from source ###
+# SPOOLES include flags (e.g. -I$(HOME)/SPOOLES.2.2 )
+SPOOLES_INCLUDE   = -I/usr/include/spooles/
+# SPOOLES library flags (e.g. $(HOME)/SPOOLES.2.2/spooles.a)
+SPOOLES_LIBS      = -lspooles
+#
+# ARPACK include flags (e.g. -I$(HOME)/ARPACK)
+ARPACK_INCLUDE    =
```
Eder-K, line 14: Should this line be commented out?

MakisH: I also thought so, but it does not make any difference, does it?

MakisH:
The tests are expected to fail, as we need to update the system tests. I am merging this to `develop`, @Eder-K you can now also merge in the system tests.

PR #45, opened 2020-11-07
Add -fallow-argument-mismatch to support GCC 10

MakisH:
See https://calculix.discourse.group/t/build-failures-with-gcc-10/76/

This is needed e.g. in Ubuntu 20.10 with CalculiX 2.16. This should be fixed in CalculiX 2.17.

@fsimonis I assume this is what you meant in the above thread?
@KyleDavisSA please check if this small change works in your case and merge if you agree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ else
 	CC = mpicc
 endif
 
-FFLAGS = -Wall -O3 -fopenmp $(INCLUDES)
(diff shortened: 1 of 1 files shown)


MakisH:
I am merging this to move forward. It works at least on my system and it looks like we need the same change for CalculiX 2.17.

fsimonis:
> @fsimonis I assume this is what you meant in the above thread?

@MakisH Yes exactly. Sorry for the late reply :+1:

PR #47, opened 2020-11-17
Remove -fallow-argument-mismatch (revert #45)

MakisH:
In #45, I added `-fallow-argument-mismatch` to make the adapter compile also with GCC10 or newer, which is more picky. However, in older compilers (such as on Ubuntu 18.04), this option does not exist and `make` fails.

Until we get a proper solution, I propose to revert this and mention in in the documentation. I moved the additional flag to a comment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,8 @@ else
 	CC = mpicc
 endif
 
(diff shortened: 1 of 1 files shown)


MakisH:
I added a note in the documentation: https://github.com/precice/calculix-adapter/wiki/Building-the-Adapter#compiling-with-gcc-10-or-newer

PR #48, opened 2020-11-17
Bugfix: remove -fallow-argument-mismatch from Makefile (add manually for GCC 10)

MakisH:
See #47. We need this to restore compatibility with older compilers (currently prominent among our users).


MakisH:
This resolves https://github.com/precice/systemtests/issues/261

PR #49, opened 2020-12-16
Add functionality to handle 2D interfaces

IshaanDesai:
This PR adds functionality to enable the adapter to handle 2D interfaces in quasi 2D-3D scenarios. From the perspective of preCICE the adapter behaves in a 2D way, and on the CalculiX side a 3D simulation is solved. 
This functionality is necessary to the new proposed structure of tutorials described here: https://github.com/precice/tutorials/issues/112

This functionality is in development and being tested with the `FSI/flap_perp/SU2-CalculiX` tutorial on the branch: https://github.com/precice/tutorials/tree/formatSU2-CCX-tutorial

### DIFF
--- a/adapter/CCXHelpers.h
+++ b/adapter/CCXHelpers.h
@@ -155,7 +155,7 @@ void getNodeDisplacementDeltas( ITG * nodes, ITG numNodes, int dim, double * v, 
  * @param co: CalculiX array with the coordinates of all the nodes
  * @param faceCenters: output array with the face centers of the input element faces
  */
-void getTetraFaceCenters( ITG * elements, ITG * faces, ITG numElements, ITG * kon, ITG * ipkon, double * co, double * faceCenters, ITG * preciceFaceCenterIDs );
+void getTetraFaceCenters( ITG * elements, ITG * faces, ITG numElements, ITG * kon, ITG * ipkon, double * co, double * faceCenters );
 
 /**
  * @brief Gets a list of node IDs from a list of input element faces
@@ -261,7 +261,7 @@ void setNodeTemperatures( double * temperatures, ITG numNodes, int * xbounIndice
  * @param xforcIndices: indices of the xforc array to modify
  * @param xforc: CalculiX array containing the (componentwise) assigned force values
  */
-void setNodeForces( ITG * nodes, double * forces, ITG numNodes, int dim, int * xforcIndices, double * xforc );
+void setNodeForces( double * forces, ITG numNodes, int dim, int * xforcIndices, double * xforc );
 
 /**
  * @brief Modifies the values of the displacements at the interface, as a Dirichlet boundary condition
--- a/adapter/CCXHelpers.h
+++ b/adapter/CCXHelpers.h
@@ -312,8 +312,27 @@ bool isDoubleEqual(const double a, const double b);
  * @brief Set all values of an array to 0
  * @param values is the array carrying double values
  * @param length is the number of elements in array
+ * @param dim is the dimension of the array data
  */
-void setDoubleArrayZero(double * values, const int length);
+void setDoubleArrayZero(double * values, const int length, const int dim);
+
+/**
+ * @brief Maps vector data from 2D mesh nodes to 3D mesh nodes
+ * @param values2D is the array of vector values on 2D mesh nodes
+ * @param mapping2D3D is a mapping between indices of 2D mesh and 3D mesh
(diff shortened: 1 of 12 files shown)


adapter/CCXHelpers.h
```diff
@@ -306,7 +306,32 @@ bool isEqual(const char * lhs, const char * rhs);
...
+
+/**
+ * @brief Set all values of an array to 0
+ * @param values is the array carrying double values
+ * @param length is the number of elements in array
+ */
+void setDoubleArrayZero(double * values, const int length);
+
+/**
+ * @brief Maps data from 2D mesh nodes to 3D mesh nodes
```
uekerman, line 319:
```suggestion
 * @brief Maps vector-valued data from 2D mesh nodes to 3D mesh nodes
```
The CalculiX adapter also supports CHT, where we also couple scalar data (temperature). We should keep this in mind for later.

DavidSCN:
Tested it now and it doesn't work:
```bash
---[precice] ERROR:  Value index out of range. Please check that the size of Displacements is correct.
```

DavidSCN:
I needed to change the config read from plural to singular in order to follow the [new naming convention](https://www.precice.org/community-contribute-to-precice.html#contributing-tutorials). The formatting might still be wrong, there are some tabs in use ?!

DavidSCN:
@IshaanDesai here a short update: the 2d3d stuff works in case we have a 2D case and enable the NLGEOM option. If we drop the option and start using the linear solver, we obtain the known issue. I think there is a logical error in the time loop (when a checkpoint is created).
However, the precice dim=3 cases fail.

DavidSCN:
I'm nearly completely convinced: the issue is in `Precice_WriteIterationCheckpoint`. Without looking deeper into the poorly documented `SimulationData` struct, I cannot resolve it, but maybe it is now obvious enough for others.

adapter/PreciceInterface.c
```diff
@@ -530,18 +649,26 @@ void PreciceInterface_ConfigureTetraFaces( PreciceInterface * interface, Simulat
...
+  interface->node2DVectorData = malloc( interface->num2DNodes * 2 * sizeof( double ) );
+
+  int dim = interface->dim;
+  for (int i = 0; i < interface->num2DNodes; i++)
+  {
+    interface->node2DScalarData[i] = 0.0;
+    interface->node2DVectorData[i*dim] = 0.0;
+    interface->node2DVectorData[i*dim + 1] = 0.0;
+  }
+
```
DavidSCN, line 664: The 3D case fails here, since `interface->num2DNode` is uninitialized and you have essentially a memory leak. All '2D' needs to be conditional.

adapter/PreciceInterface.c
```diff
@@ -448,41 +517,91 @@ void PreciceInterface_ConfigureFaceCentersMesh( PreciceInterface * interface, Si
...
+	interface->nodeCoordinates = malloc( interface->numNodes * interface->dimCCX * sizeof( double ) );
+	getNodeCoordinates( interface->nodeIDs, interface->numNodes, interface->dimCCX, sim->co, sim->vold, sim->mt, interface->nodeCoordinates );
+
+  int count = 0;
+  int dim = interface->dim;
+  int dimCCX = interface->dimCCX;
+  // Extract 2D coordinates from 3D coordinates for quasi 2D-3D coupling
+  if( interface->quasi2D3D )
+  {
+    interface->num2DNodes = interface->numNodes / 2;
```
DavidSCN, line 548: Or move the initialization out of the if statement

IshaanDesai:
> Without looking deeper into the poorly documented `SimulationData` struct, I cannot resolve it, but maybe it is now obvious enough for others.

I will look into this.

> The formatting might still be wrong, there are some tabs in use ?!

I dont understand the question here, do you mean the indentation in the code is not correct? --> for the indentation I agree that rigorous formatting of the CalculiX-Adapter is necessary. I will open another PR after the quasi 2D-3D is done.

Thanks for debugging the 3D case! :smile:

DavidSCN:
> I dont understand the question here, do you mean the indentation in the code is not correct? --> for the indentation I agree that rigorous formatting of the CalculiX-Adapter is necessary. I will open another PR after the quasi 2D-3D is done.

I was just pointing out that my commit was formatted inconsistently with the rest of the code.

DavidSCN (approved):
I tested it now using ubuntu 20.10 and 20.04 in combination with the new restructure and it works fine.

PR #50, opened 2021-02-12
Code refinement

IshaanDesai:
This PR makes the 2D-3D implementation more readable and reverts a bug introduced in the PR: https://github.com/precice/calculix-adapter/pull/49
The bug was observed as the 3D_Tube tutorial case with OpenFOAM-CalculiX was failing. One major task:
- [x] Ensure the 3D_Tube tutorial case works for the full duration. @DavidSCN can you please check this to rule out system specific behavior on my side?
- [x] Ensure all FSI tutorial cases with CalculiX run and the results are reasonable

### DIFF
--- a/adapter/CCXHelpers.c
+++ b/adapter/CCXHelpers.c
@@ -535,6 +535,11 @@ bool isDoubleEqual(const double a, const double b)
 	return fabs(a - b) < 1.e-14;
 }
 
+bool isQuasi2D3D(const int quasi2D3D)
+{
+	return quasi2D3D == 1;
+}
+
 void setDoubleArrayZero(double * values, const int length, const int dim)
 {
 	ITG i, j;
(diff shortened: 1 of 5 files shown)


IshaanDesai:
Running the perpendicular flap case (CalculiX - OpenFOAM) in the new restructured tutorials format here: https://github.com/precice/tutorials/pull/146 produces the result:
image
which is reasonable and close to the reference result.

DavidSCN (approved):
Works much better than before, thanks. I tested it now until t=0.005s for the 3D tube and I can confirm that I obtain exactly the same result as the `master` branch. However, there is still something wrong with the current master branch. Let me open up an issue on this topic.

DavidSCN:
...also tested the 3D until the end (n_t=100)... works (ignoring #51 ).

PR #68, opened 2021-09-12
Merge master into v2.16

KyleDavisSA:
This PR merges the current master branch into branch v2.16. 

Recently many changes have been made to the master branch, which currently supports CalculiX v2.16. The branch v2.16 must always remain up-to-date with the latest developments for CalculiX v2.16. The latest version of CalculiX is v2.17. Therefore, the master branch should support v2.17. This requires first that the master branch is merged into v2.16 to keep v2.16 adapter up-to-date. 

A follow-up PR will be made to merge the new changes into v2.17.


KyleDavisSA:
I currently use the master branch and it all works as it should for me.

IshaanDesai (approved):
Makes sense to keep the `v2.16` branch synced with `master`

MakisH (approved):
This works fine for me with CalculiX 2.16. This update is also important so that CalculiX 2.16 can work with our 2D tutorials and so that the building instructions correspond to the documentation.

By the way, since #41 (I guess), I cannot directly build the adapter with CalculiX built from source in its own folder and not installed (my usual development environment). I needed to add:

```diff
LIBS = \
	$(SPOOLES_LIBS) \
	$(PKGCONF_LIBS) \
+	-lprecice \
	-lstdc++ \
```

If more of you have a similar issue, maybe we should add it anyway.

IshaanDesai:
@MakisH 

> If more of you have a similar issue, maybe we should add it anyway.

I dont need to add `-lprecice` to `LIBS` for successful compilation. Even if CalculiX is installed from source in its own folder you can just add the `CCX` path in the Makefile for it to be found right?

MakisH:
> I dont need to add `-lprecice` to `LIBS` for successful compilation. Even if CalculiX is installed from source in its own folder you can just add the `CCX` path in the Makefile for it to be found right?

I am talking about preCICE here, not CalculiX. This is just a problem we need to fix in the pkg-config of preCICE, but it is rather complicated. But maybe it only affects me.

IshaanDesai:
> I am talking about preCICE here, not CalculiX. This is just a problem we need to fix in the pkg-config of preCICE, but it is rather complicated. But maybe it only affects me.

Okay I misunderstood you previously :sweat_smile: still if anything needs to be tested before standardizing the Makefile then I can test it.

KyleDavisSA:
But it seems that adding `-lprecice` solves the problem when preCICE is installed in its own directory? I don't think this impacts merging this PR?

MakisH:
> But it seems that adding `-lprecice` solves the problem when preCICE is installed in its own directory? I don't think this impacts merging this PR?

It does not impact this PR at all. I just noticed it while reviewing this one. I should have asked on Gitter / open an issue.

Issue #57, opened 2021-04-07
CalculiX + precice with PaStiX solver (pastix.c and pastix.h)


3rav:
**Attempting to compile CalculiX + precice adapter with PaStiX solver ends with an error:**

```
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:18:35: error: unknown type name 'ITG'
   18 |          double *sigma,double *b, ITG *icol, ITG *irow,
      |                                   ^~~
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:18:46: error: unknown type name 'ITG'
   18 |          double *sigma,double *b, ITG *icol, ITG *irow,
      |                                              ^~~
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:19:3: error: unknown type name 'ITG'
   19 |   ITG *neq, ITG *nzs,ITG *symmetryflag,ITG *inputformat,ITG *jq,
      |   ^~~
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:19:13: error: unknown type name 'ITG'
   19 |   ITG *neq, ITG *nzs,ITG *symmetryflag,ITG *inputformat,ITG *jq,
      |             ^~~
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:19:22: error: unknown type name 'ITG'
   19 |   ITG *neq, ITG *nzs,ITG *symmetryflag,ITG *inputformat,ITG *jq,
      |                      ^~~
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:19:40: error: unknown type name 'ITG'
   19 |   ITG *neq, ITG *nzs,ITG *symmetryflag,ITG *inputformat,ITG *jq,
      |                                        ^~~
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:19:57: error: unknown type name 'ITG'
   19 |   ITG *neq, ITG *nzs,ITG *symmetryflag,ITG *inputformat,ITG *jq,
      |                                                         ^~~
C:/msys64/usr/local/CalculiX/ccx_2.17_precice2/src/pastix.h:20:3: error: unknown type name 'ITG'
   20 |   ITG *nzs3,ITG *nrhs);
      |   ^~~
[...]

PR #71, opened 2021-10-09
Fixed CCXhelpers for 8 bytes + added helper files for PaStiX

boris-martin:
(See related documentation PR : https://github.com/precice/precice.github.io/pull/93 )

- Check inconsistent types in CCXHelpers, where some ITG were int. This only breaks when using -DLONGLONG, which explains why it was never spotted earlier. (This option is required when using PaStiX). See https://github.com/precice/calculix-adapter/issues/57
- Added a PaStiX-specific makefile and a script to make the code PaStiX-compatible (see again the same issue, with naming conflicts in the headers)

Closes https://github.com/precice/calculix-adapter/issues/57


Makefile_i8_PaStiX
```diff
@@ -0,0 +1,133 @@
+# See our wiki for getting the CalculiX dependencies:
```
MakisH, line 1: Maybe you want to base this `Makefile` to the one from `master` as well?
Also, look out for any references to the wiki, which now need to point to the adapter documentation on the website.

Makefile_i8_PaStiX
```diff
@@ -0,0 +1,133 @@
...
+OBJDIR = bin
+
+# For PaStiX only
+PASTIX_LIBS = $(HOME)/PaStiX/hwloc_i8/lib/libhwloc.so $(HOME)/OpenBLAS_i8/libopenblas.a \
+  $(HOME)/PaStiX/pastix_i8/lib/libpastix.a $(HOME)/OpenBLAS_i8/libopenblas.a -lpthread -lm $(HOME)/PaStiX/pastix_i8/lib/libspm.a \
+  $(HOME)/PaStiX/pastix_i8/lib/libpastix_parsec.a $(HOME)/PaStiX/pastix_i8/lib/libpastix_kernels.a $(HOME)/OpenBLAS_i8/libopenblas.a -lrt \
+  $(HOME)/PaStiX/pastix_i8/lib/libpastix_kernels_cuda.a $(HOME)/PaStiX/parsec_i8/lib/libparsec.so \
+  /usr/lib/x86_64-linux-gnu/libcudart_static.a -lpthread -ldl -lrt /usr/lib/x86_64-linux-gnu/libcublas_static.a \
+  /usr/lib/x86_64-linux-gnu/libculibos.a /usr/lib/x86_64-linux-gnu/libcublasLt_static.a -lrt \
+  $(HOME)/PaStiX/scotch_i8/lib/libscotch.a $(HOME)/PaStiX/scotch_i8/lib/libscotcherrexit.a -lpthread -lz -lm /usr/lib/x86_64-linux-gnu/libstdc++.so.6 /usr/lib/x86_64-linux-gnu/libgomp.so.1
```
MakisH, line 41: These paths sound very platform-specific. Maybe at least set this path prefix to a variable somewhere?

Makefile_i8_PaStiX
boris-martin: I'm not sure I follow, this is already a modified version of `master` makefile. 
Agree for the wiki, I didn't really pay attention to this at first.

Makefile_i8_PaStiX
boris-martin: What do you mean by prefix ?
I could split this into one variable per library though.

MakisH: By prefix, I mainly mean the `/usr/lib/x86_64-linux-gnu/`.

Makefile_i8_PaStiX
MakisH: Ah, right, I was confused for a moment.

KyleDavisSA:
The issue with the ITG types, can we merge this separately to any pastixs changes?

boris-martin:
> The issue with the ITG types, can we merge this separately to any pastixs changes?

Definitely, although it's gonna mess up the PRs a bit and no one complained until PaStiX came

MakisH:
> 

But this would mean that we split this into two: one for the ` adapter/CCXHelpers[.c|.h]` and one for the ` Makefile_i8_PaStiX` and `pastix_pre_build.sh`.

Makefile_i8_PaStiX
MakisH: Keeping this unresolved only to remember to update according to the latest `Makefile`.

By the way, instead of maintaining two Makefiles, could we maybe add a separate target in the same Makefile?

boris-martin: Probably, I can look into that

Makefile_i8_PaStiX
```diff
@@ -0,0 +1,131 @@
...
+# Get the CFLAGS and LIBS from pkg-config (preCICE version >= 1.4.0).
+# If pkg-config cannot find the libprecice.pc meta-file, you may need to set the
+# path where this is stored into PKG_CONFIG_PATH when building the adapter.
+PKGCONF_CFLAGS  = $(shell pkg-config --cflags libprecice)
+PKGCONF_LIBS    = $(shell pkg-config --libs libprecice)
+
+# Specify where to store the generated .o files
+OBJDIR = bin
+
+# To get the correct path, export the required folders for pkg-config. Example: export PKG_CONFIG_PATH=~/PaStiX/pastix_i8/lib/pkgconfig/:~/PaStiX/hwloc_i8/lib/pkgconfig/:~/PaStiX/parsec_i8/lib/pkgconfig/
```
MakisH, line 33: Don't forget to *extend* such environment variables, not *overwrite* them. The user may already have other paths defined in their `PKG_CONFIG_PATH`.

Makefile_i8_PaStiX
```diff
@@ -0,0 +1,131 @@
...
+
+PASTIX_INCLUDE = $(shell pkg-config --cflags pastix)
+PASTIX_LIBS = $(shell pkg-config --libs pastix hwloc parsec lapacke) $(HOME)/OpenBLAS_i8/libopenblas.a \
```
MakisH, line 36: Is this really `lapacke`? Feels like a typo.

Makefile_i8_PaStiX
```diff
@@ -0,0 +1,131 @@
...
+$(HOME)/PaStiX/pastix_i8/lib/libpastix_kernels.a \
```
MakisH, line 37: Since you are now reusing the hard-coded `$(HOME)` further down in this file, maybe we should define a variable at the top of the file.

Makefile_i8_PaStiX
```diff
@@ -0,0 +1,131 @@
...
+   /usr/lib/x86_64-linux-gnu/libcublas_static.a \
+  /usr/lib/x86_64-linux-gnu/libculibos.a /usr/lib/x86_64-linux-gnu/libcublasLt_static.a \
+  $(HOME)/PaStiX/scotch_i8/lib/libscotch.a $(HOME)/PaStiX/scotch_i8/lib/libscotcherrexit.a
```
MakisH, line 40: I understand that you are specifying all the dependencies of pastix here, because you are using pastix as a static library. This should be fine, but I feel it would be simpler to use pastix as a shared library, so that we don't need to specify all these. Is pastix available as a shared library? If not, can we get all these directly from the pkg-config of pastix?

Makefile_i8_PaStiX
boris-martin: Apparently `lapacke` is a C wrapper of `lapack` (written in FORTRAN) or something like that. But now that you say it, I tried `lapack` and it worked too... Because none of them were required.
I think the required dependency is contained in `OpenBLAS_i8/libopenblas.a`. So let's just remove it alltogether

Makefile_i8_PaStiX
boris-martin: What do you mean ? Something like "PASTIX_DEP_FOLDER" ?
My idea was that if you don't do it on home, maybe all the other assumptions (subdirectory PaStiX etc) are wrong, se the user would have to change all paths anyways

Makefile_i8_PaStiX
boris-martin: I already use the pkg-config for pastix, but it doesn't contain the `libpastix_kernels.a`, surprisingly.
As for shared PaStiX... Since we use a specific version of PaStiX designed for CalculiX I wouldn't trust the possibilities, since CalculiX gives very specific instructions.
However the CMakeFile of pastix does contain a SHARED mode, so I can give it a try.

Makefile_i8_PaStiX
MakisH: Actually, what does the official CalculiX Makefile do regarding PaStiX? I hope I am not going way back to the basics.

We could treat it similarly to SPOOLES and ARPACK and define `PASTIX_INCLUDE` and `PASTIX_LIBS`.

Makefile_i8_PaStiX
MakisH: If this is indeed a special version, then let's keep it static to avoid any accidental linking to the wrong one.

MakisH:
@boris-martin could you please also resolve any open conversations here? Let me know when you need another review.

PR #72, opened 2021-10-21
Automatic creation of .deb packages

boris-martin:
This adds a Github Actions workflow that download dependencies and Calculix source code, builds the adapter (the "regular" version, without PaStiX), creates a .deb package and uploads it as artifact. (It can then be downloaded from Github on the workflow run)
Currently runs on Ubuntu 18.04 and 20.04 (The 2 latest LTS versions). It should work on other versions but I couldn't get Github to run on 21.04, and it stalled the entire workflow forever so I removed it.

Usage : download the corresponding artifact and `sudo apt install [packagefile]`. This requires an installation of `preCICE` through the provided packages (i.e. this package won't work with a custom build of preCICE, as it explicitly requires the libprecice package)

The makefile has been updated to be similar (just replaced 2.16 by 2.17 everywhere) to the one of the master branch instead of  the current v2.17, which is outdated and painful to use, especially within scripts.

Things yet to do : 

- [x] Write some stuff in the man page. There is a script that converts if from Markdown to the appropriate format, so the file to change is `packaging/manpage.md`
- [x] Write a correct copyright file. Still a bit confused as the adapter is GPL3 and CalculiX seems to be GPL2, and while the copyright mechanism allows different files to have different licenses, the binary is a mix of both. Insights appreciated ! 
**FIXED : CalculiX says "distribute with GPL2 or above" which solves the problem. GPL3 it is.**
- [x] Currently, the script builds with preCICE 2.3 but the configuration of the package requires >= 2.2, which is inconsistent. I don't know in what direction to settle. Should we link to an outdated preCICE version, or force people to upgrade ?
**FIX : Package indicates 2.0.0 is enough.**

Things to do or think about in the long run :
- The script is v2.17-specific, but could be made more flexible and build for different versions etc.
- The package has version "2.17-1" and it's hardcoded, I don't know if we should number the versions according to CCX or to preCICE.
- I know the preCICE packages has security stuff like a sha256 key, but I have zero knowledge about it, so currently the package doesn't feature this. 
- Building the adapter with PaStiX is much harder, and I spent too many hours trying to do it automatically. (For some reasons, I miss some CUDA related dependencies on Github Actions despite installing the required package, though I have no issues with my own VMs). Furthermore, PaStiX requires a lot of hand-built libraries (with specific flags and so on), so integrating it with the Ubuntu ecosystem is impossible, and we'd have to package the dependencies too, and... Well, it looks like a nightmare.
- Once the packages are made, we still have to release them.


fsimonis:
The following is a good read regarding the license compatibility: https://www.gnu.org/licenses/license-compatibility.html

CalculiX is licensed under the GPLv2 license, which requires any redistributed versions to be licensed under **the same license**. So, we have an issue.

boris-martin:
> The following is a good read regarding the license compatibility: https://www.gnu.org/licenses/license-compatibility.html
> 
> CalculiX is licensed under the GPLv2 license, which requires any redistributed versions to be licensed under **the same license**. So, we have an issue.

Can't we simply have the adapter be GPL2 instead of GPL3 ? That looks like the best way to remove headaches to me. Unless we can't because we'd need the authorization of all people who contributed to the adapter or something like this ?

Makefile
```diff
@@ -1,14 +1,24 @@
 # See our wiki for getting the CalculiX dependencies:
```
MakisH, line 1: You could also merge `master` to `v2.17` or cherry-pick the respective commit, to avoid this duplicate change.

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04]
+        include:
+          - os: ubuntu-18.04
+            name: bionic
+          - os: ubuntu-20.04
+            name: focal
+#          - os: ubuntu-21.04
+#            name: hirsute
```
MakisH, line 17: You can remove this. In any case, 21.04 will soon reach end-of-life. If you want to test against a non-LTS, try 21.10.

```suggestion
```

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+    runs-on: ${{matrix.os}}
+    continue-on-error: true
+
+    steps:
+    - name: Update system
+      run: sudo apt update && sudo apt upgrade -y
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.ref }}
+    - name: get preCICE
```
MakisH, line 27: Tiny consistency thing:
```suggestion
    - name: Get preCICE
```
(also below)

To make it clearer for the future, I would include the preCICE version in the name:

```suggestion
    - name: Get preCICE v2.3.0
```

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+      run: wget https://github.com/precice/precice/releases/download/v2.3.0/libprecice2_2.3.0_${{matrix.name}}.deb &&
+        sudo apt install ./libprecice2_2.3.0_${{matrix.name}}.deb -y
+    - name: install dependencies
+      run: sudo apt install libarpack2-dev libspooles-dev libyaml-cpp-dev -y
+    - name: Download Calculix
```
MakisH, line 32:
```suggestion
    - name: Download CalculiX
```
(also below)

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+      run: wget http://www.dhondt.de/ccx_2.17.src.tar.bz2
+    - name: Unpack Calculix
+      run: bzip2 -d ccx_2.17.src.tar.bz2 && tar -xf ccx_2.17.src.tar
```
MakisH, line 35: Why not directly
```suggestion
      run: tar -xjf ccx_2.17.src.tar.bz2
```
?

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+    - run: echo "CCX=$(pwd)/CalculiX/ccx_2.17/src" >> $GITHUB_ENV
```
MakisH, line 36: Missing a name for the step, which could also be merged with the next one.

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Show paths
+      run: echo "${{env.CCX}}" && ls "${{env.CCX}}"
+    - name: make
+      run: make -j 4 CCX=${CCX}
```
MakisH, line 40: Why before `${{env.CCX}}` but here `${CCX}`? Could this lead to a bug in some situation?

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Store Debian package artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: "calculix-precice_2.17-1_amd64_${{matrix.name}}.deb"
+        path: "packaging/calculix-precice_2.17-1_amd64_${{matrix.name}}.deb"
+    # - name: Store Debian Package
+    #   uses: actions/upload-artifact@v2
+    #   with:
+    #     name: calculix-precice_2.17-1_amd64.deb
+    #     path: ./packaging/calculix-precice_2.17-1_amd64.deb
```
MakisH, line 64: This can be removed.

.github/workflows/ubuntu_build.yml
```diff
@@ -0,0 +1,66 @@
...
+    # - name: Inspect package with lintian
+    #   run: sudo apt install lintian -y && lintian ./packaging/calculix-precice_2.17-1_amd64.deb
```
MakisH, line 66: This should be enabled eventually.

packaging/README.md
```diff
@@ -0,0 +1,10 @@
+# Automatic packaging of the adapter
+
+This contains the necessary files to build a Debian package containing the adapter. (Built without PaStiX)
+
+## Note for maintainers 
+
+- Keep up to date the list of dependencies in the control file with `dpkg-shlibs`. Care must be taken when preCICE gets updated, typically.
+- Update the changelog.
+- The script `make_deb.sh` does the required actions to compress the changelog and manpage and put them at the right place. Building the package should be as simple as running it.
+- `lintian` and `pandoc` are required to run that script. The first checks the validity of the package and the second compiles the `man` page from Markdown into the appropraite format
```
MakisH, line 10:
```suggestion
- The script `make_deb.sh` compresses the changelog and manpage and puts them at the right places. Building the package should be as simple as running this script.
    - Dependencies: `lintian` (to validate the package) and `pandoc` (to generate the `man` page).
```
What I did here:
- I tried to make it more concise (deliver the same content in fewer words, more accurately)
- Replaced "running it" and "run that script" with more specific formulations. This is helpful when skimming through documentation.

packaging/calculix-precice_2.17-1_amd64/DEBIAN/control
```diff
@@ -0,0 +1,14 @@
+Package: calculix-precice
+Version: 2.17-1
+Source: calculix-precice
+Architecture: amd64
+Section: contrib/science
+Priority: optional
+Maintainer: The preCICE team (precice.org) <info@precice.org>
+Depends: libarpack2 (>= 2.1), libblas3 | libblas.so.3, libc6 (>= 2.29), libgcc-s1 (>= 4.0), libgfortran5 (>= 8), libgomp1 (>= 4.9), liblapack3 | liblapack.so.3, libprecice2 (>= 2.2.0), libspooles2.2, libstdc++6 (>= 9), libyaml-cpp0.6 (>= 0.6.2)
```
MakisH, line 8: As you also mentioned, we are building with preCICE v2.3.0, but requiring v2.2.0. Is there a specific reason for v2.2.0? In principle, if we are applying semantic versioning correctly, v2.0.0 should be the minimum required.

```suggestion
Depends: libarpack2 (>= 2.1), libblas3 | libblas.so.3, libc6 (>= 2.29), libgcc-s1 (>= 4.0), libgfortran5 (>= 8), libgomp1 (>= 4.9), liblapack3 | liblapack.so.3, libprecice2 (>= 2.0.0), libspooles2.2, libstdc++6 (>= 9), libyaml-cpp0.6 (>= 0.6.2)
```

packaging/calculix-precice_2.17-1_amd64/DEBIAN/control
```diff
@@ -0,0 +1,14 @@
...
+Bugs: https://github.com/precice/calculix-adapter/issues
+Description: PreCICE adapter for CalculiX.
```
MakisH, line 10: Maybe, to avoid the change in style:
```suggestion
Description: The official preCICE adapter for CalculiX.
```

(if linter complains for `Description: preCICE adapter for CalculIX.`)

packaging/calculix-precice_2.17-1_amd64/usr/share/doc/calculix-precice/copyright
```diff
@@ -0,0 +1,9 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: preCICE
+Upstream-Contact: preCICE mailing-list <info@precice.org>
```
MakisH, line 3:
```suggestion
Upstream-Contact: preCICE developers <info@precice.org>
```

packaging/calculix-precice_2.17-1_amd64/usr/share/doc/calculix-precice/copyright
```diff
@@ -0,0 +1,9 @@
...
+Copyright: (c) 2019 the preCICE developers <info@precice.org>
+Source: https://github.com/precice/precice
+
+Files: *
+Copyright: (c) 2019 the preCICE developers <info@precice.org>
```
MakisH, line 8: Now I am wondering if we should update the date every now and then.
We should also mention somewhere the license of CalculiX.

packaging/calculix-precice_2.17-1_amd64/usr/share/doc/calculix-precice/copyright
```diff
@@ -0,0 +1,9 @@
...
+License: LGPL-3
```
MakisH, line 9: Why **L**GPL? This is the license of preCICE, the adapter currently is the regular GPL.

packaging/make_deb.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/bash
+
+DISTRIBUTION="_$1"
+echo $DISTRIBUTION
+
+# Compress the changelog, strip the binaries
+cp changelog.Debian calculix-precice_2.17-1_amd64/usr/share/doc/calculix-precice/changelog.Debian
+# Options : -f for removing file if it was there, -n for no time stamp
+gzip -9 calculix-precice_2.17-1_amd64/usr/share/doc/calculix-precice/changelog.Debian -f -n
```
MakisH, line 9: I did not know what `-9` is for (best compression). Maybe use `--best`, instead, as it is a bit more intuitive?
```suggestion
# Options : --best for best compression, -f for removing file if it was there, -n for no time stamp
gzip --best -f -n calculix-precice_2.17-1_amd64/usr/share/doc/calculix-precice/changelog.Debian
```

(same below)

packaging/make_deb.sh
```diff
@@ -0,0 +1,26 @@
...
+
+pandoc manpage.md -s -t man -o ccx_preCICE.1
+mkdir -p calculix-precice_2.17-1_amd64/usr/share/man/man1
+gzip -9 -n -f ccx_preCICE.1
+chmod 644 ccx_preCICE.1.gz
+mv ccx_preCICE.1.gz calculix-precice_2.17-1_amd64/usr/share/man/man1
+
+# Copy to a folder with appropriate postfix
+
+cp -r calculix-precice_2.17-1_amd64/ "calculix-precice_2.17-1_amd64$DISTRIBUTION"
```
MakisH, line 22: I don't really like that we have the versions hardcoded everywhere. But I also understand that it is complicated to pass this information everywhere needed.

Maybe add some documentation in the README on where one needs to update the version and what the versioning scheme is?

packaging/manpage.md
```diff
@@ -0,0 +1,39 @@
...
+# NAME
+ccx_preCICE - Adapter for using CalculiX with the preCICE library.
+
+# SYNOPSIS
+**ccx_preCICE** [*OPTION*] *files*
+
+# DESCRIPTION
+
+This page briefly summarizes how to use CalculiX with or without preCICE. For more detailed instructions, read the CalculiX documention on http://www.dhondt.de/ccx_2.17.pdf and preCICE documentation on https://precice.org/docs.html.
+CalculiX is used alone when the -precice-participant option is not used.
```
MakisH, line 14:
```suggestion
Omitting the -precice-participant option starts a stand-alone CalculiX simulation.
```

(but maybe move this further down, to the options section?)

packaging/manpage.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Using with preCICE
+
+## Regular usage
```
MakisH, line 18: What is the purpose of these sections and why are they here?

packaging/manpage.md
```diff
@@ -0,0 +1,39 @@
...
+
+: Starts the job defined in <jobname>.inp. Be sure to omit the extension .inp in the job name.
+
+**-precice-participant < participant >**
+
+: Runs a preCICE simulation with name < participant >. This name must be the one used in the YAML config file as well as in the preCICE configuration file.
+
+# EXAMPLES
+
+**ccx_preCICE -i flap -precice-participant Calculix** : Runs the coupled simulation flap.inp as participant "CalculiX".
```
MakisH, line 37:
```suggestion
**ccx_preCICE -i flap -precice-participant CalculiX** : Runs the coupled simulation flap.inp as participant "CalculiX".
```

MakisH (changes requested):
Very good so far! I already commented on tiny details, thus the amount of comments. Let's have another review iteration later, as now I only looked at the code.

Makefile
boris-martin: Yes, ideally `v2.17` should have been fully updated before we merge this. I asked @KyleDavisSA about it but I'm not confident to do that merge myself.

.github/workflows/ubuntu_build.yml
boris-martin: Never heard of it, good to know !

Makefile
MakisH: Why not? Let me know if I can help.

.github/workflows/ubuntu_build.yml
MakisH: :grimacing: https://precice.org/adapter-calculix-get-calculix.html#get-the-source

packaging/manpage.md
boris-martin: I planned to add the part above, but since it's very short these sections should probably be removed

boris-martin:
> > Currently runs on Ubuntu 18.04 and 20.04 (The 2 latest LTS versions). It should work on other versions but I couldn't get Github to run on 21.04, and it stalled the entire workflow forever so I removed it.
> 
> Where did it hang with 21.04? Did you get any error messages?

No errors, just infinite waiting. Maybe they have already switched everything to 21.10 ?

> > Usage : download the corresponding artifact and `sudo apt install [packagefile]`. This requires an installation of `preCICE` through the provided packages (i.e. this package won't work with a custom build of preCICE, as it explicitly requires the libprecice package)
> 
> We would then also need to add this information to the adapter documentation.

Yes, I was thinking of waiting for the official release.

> > The makefile has been updated to be similar (just replaced 2.16 by 2.17 everywhere) to the one of the master branch instead of the current v2.17, which is outdated and painful to use, especially within scripts.
> 
> Can we just merge this change everywhere needed, please? :-) This is anyway what we have currently documented.

I'm all in favor, but I'd prefer someone more experienced handled that. (Not really changing the makefile, but 2.17 branch is already 45 commits behind master, so that's slightly messy)

> > ```
> > * The package has version "2.17-1" and it's hardcoded, I don't know if we should number the versions according to CCX or to preCICE.
> > ```
> 
> We need a versioning scheme here and we should put some thought on it. Since the adapter works with only specific versions of CalculiX, but any preCICE v2.x, I would suggest something like `calculix-precice2-2.17-1`.

By the way, I'm not 100% sure what preCICE version to use for the build. If we compile with 2.3 but use only features from <= 2.2, would the adapter work on a computer with preCICE 2.2 installed, for instance ? More generally, should we always build with the most recent version of preCICE 2.x ?

> > ```
> > * I know the preCICE packages has security stuff like a sha256 key, but I have zero knowledge about it, so currently the package doesn't feature this.
> > ```
> 
> You can generate such a hash very easily, with `sha256sum <file>`. This is mainly to certify that the package has not been modified.

So we simply need to provide the output of sha256sum so that people can check ? 

> > ```
> > * Building the adapter with PaStiX is much harder, and I spent too many hours trying to do it automatically. (For some reasons, I miss some CUDA related dependencies on Github Actions despite installing the required package, though I have no issues with my own VMs). Furthermore, PaStiX requires a lot of hand-built libraries (with specific flags and so on), so integrating it with the Ubuntu ecosystem is impossible, and we'd have to package the dependencies too, and... Well, it looks like a nightmare.
> > ```

> 
> Let's forget about PaStiX in the Debian package right now. This PR is already a significant step ahead.

That seems very reasonable to me.

> > ```
> > * Once the packages are made, we still have to release them.
> > ```
> 
> We should eventually also start making releases here, similarly to other adapters. This would be much better than the current branching model.

Even with clean releases, there's still the issue of having different code for each CalculiX version no ?

MakisH:
> No errors, just infinite waiting. Maybe they have already switched everything to 21.10 ?

21.04 is still supported (and I am still using it).

> By the way, I'm not 100% sure what preCICE version to use for the build. If we compile with 2.3 but use only features from <= 2.2, would the adapter work on a computer with preCICE 2.2 installed, for instance ? More generally, should we always build with the most recent version of preCICE 2.x ?

We should build/test with the latest version of preCICE. But, because we are linking to a shared library, any preCICE v2.x version will work. (there was even an AdvProg slide on lecture "Build time" regarding exactly this example :sweat_smile: )

> So we simply need to provide the output of sha256sum so that people can check ?

Exactly. Unless linter suggests more steps.

> Even with clean releases, there's still the issue of having different code for each CalculiX version no ?

Yes. But we should have a clear policy on which CalculiX versions we support. I would vote to only support the latest one.

packaging/make_deb.sh
boris-martin: I had added a file "README for maintainers" at some point but it seems I lost it / forgot to add it to Git. I'll rewrite that.

boris-martin:
> > No errors, just infinite waiting. Maybe they have already switched everything to 21.10 ?
> 
> 21.04 is still supported (and I am still using it).

Weird, I should give it another try at some point.

> > By the way, I'm not 100% sure what preCICE version to use for the build. If we compile with 2.3 but use only features from <= 2.2, would the adapter work on a computer with preCICE 2.2 installed, for instance ? More generally, should we always build with the most recent version of preCICE 2.x ?
> 
> We should build/test with the latest version of preCICE. But, because we are linking to a shared library, any preCICE v2.x version will work. (there was even an AdvProg slide on lecture "Build time" regarding exactly this example 😅 )

Woops :D

> > So we simply need to provide the output of sha256sum so that people can check ?
> 
> Exactly. Unless linter suggests more steps.

As far as I remember it didn't complain about the lack of sha256, I was more thinking about how we just publish keys every time there's a new preCICE release.
> 
> > Even with clean releases, there's still the issue of having different code for each CalculiX version no ?
> 
> Yes. But we should have a clear policy on which CalculiX versions we support. I would vote to only support the latest one.

Probably a good idea to add support for 2.18 before dropping the others then !

packaging/manpage.md
```diff
@@ -0,0 +1,39 @@
+% ccx_preCICE(1) 2.17
+% Boris Martin
+% Oct 2021
```
fsimonis, line 3: How is this information processed?
The official guideline to writing man pages state that one should avoid the author block.

boris-martin: It was part of an unofficial template I used, I didn't really pay attention to.
Output on my machine is this : 

image

No objection te removing it if you think it's better.

boris-martin:
Slightly unrelated, but following this https://github.com/precice/precice/pull/1132#discussion_r746697201 I realized it was probably interesting (in the long run) to modify Makefiles to provice a `make instal` which includes a few things from here, like `man` pages, copyright and changelog.

MakisH:
> Slightly unrelated, but following this [precice/precice#1132 (comment)](https://github.com/precice/precice/pull/1132#discussion_r746697201) I realized it was probably interesting (in the long run) to modify Makefiles to provice a `make instal` which includes a few things from here, like `man` pages, copyright and changelog.

I am not sure if investing on Make would be the right thing to do in that direction. [Migrating to CMake](https://github.com/precice/calculix-adapter/pull/40) could open such possibilities, but I guess it would be complicated to do it in the downstream project (adapter) without already having support from the upstream (CalculiX + dependencies).

MakisH:
Regarding licensing, this is the license of CalculiX, as found in the `CalculiX.h` of version 2.17:

```
/*     CALCULIX - A 3-dimensional finite element program                 */
/*              Copyright (C) 1998-2020 Guido Dhondt                     */

/*     This program is free software; you can redistribute it and/or     */
/*     modify it under the terms of the GNU General Public License as    */
/*     published by the Free Software Foundation; either version 2 of    */
/*     the License, or (at your option) any later version.               */

/*     This program is distributed in the hope that it will be useful,   */
/*     but WITHOUT ANY WARRANTY; without even the implied warranty of    */ 
/*     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the      */
/*     GNU General Public License for more details.                      */

/*     You should have received a copy of the GNU General Public License */
/*     along with this program; if not, write to the Free Software       */
/*     Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.         */
```

Since we have the `either version 2 of the License, or (at your option) any later version.`, I would say we have no problem with GPLv3.

From the [FSF article](https://www.gnu.org/licenses/license-compatibility.html) that @fsimonis posted above:

> The FSF uses the approach of asking people to release programs under “GNU GPL version N or any later version.” This licensing is compatible with version N, and also with N+1 (because it offers version N+1 as an option). When you combine code under “GPL 3 or later” with code under “GPL 2 or later”, the license of the combination is their intersection, which is “GPL 3 or later”.

The correct license of `calculix-precice`, as a code derived from CalculiX, and given our additions as GPLv3, is GPLv3.

In our case, we maintain the GPLv2 notice in `CalculiX.h`, but also provide GPLv3 for our code. I think this is ok but again, I barely understand licensing.

boris-martin:
I did some improvements : 
- I finally modified the copyright file, with GPL3
- I put preCICE 2.0.0 as prerequisite instead of 2.2
- I added SHA256 key generation & export
- I cleaned a few things, especially the workflow file, with an environment variable to have less things to change when we change CalculiX version
- The package is now named calculix-precice2 instead of calculix-precice

I'm undrafting the PR, though I also plan to add a "README for maintainers".

.github/workflows/ubuntu_build.yml
boris-martin: I re-enabled this then remembered I commented it out because `lintian` is called by the `make_deb.sh` script, si it's redundant. I don't know which one to prefer : doing it in the workflow might make debugging Actions easier (if everything builds but lintian complains, just check this part of the workflow), but doing in the shell script makes the check integrated in the case we build the package locally.

packaging/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This contains the necessary files to build a Debian package containing the adapter. (Built without PaStiX)
+
+## Note for maintainers 
+
+- Keep up to date the list of dependencies in the control file with `dpkg-shlibs`. In theory, it will detect preCICE with the version used for the build, so we update it manually to the lowest possible version. For instance, if the adapter is built with preCICE 2.3.0, we update the dependency as >= 2.0.0
+- Update the changelog.
+- The script `make_deb.sh` compresses the changelog and manpage and puts them at the right places. Building the package should be as simple as running this script.
+    - Dependencies: `lintian` (to validate the package) and `pandoc` (to generate the `man` page).
+- To update CalculiX version : update the Github action workflow file (in `.github/workflows`). Update the `PACKAGE_NAME` variable, the paths (e.g. replace `wget http://www.dhondt.de/ccx_2.17.src.tar.bz2` by `wget http://www.dhondt.de/ccx_2.XX.src.tar.bz2` with XX the corresponding minor version of CalculiX and folder paths.
```
MakisH, line 11: :smirk: 

```suggestion
- To update CalculiX version: update the Github action workflow file (in `.github/workflows`). Update the `PACKAGE_NAME` variable, the paths (e.g. replace `wget http://www.dhondt.de/ccx_2.17.src.tar.bz2` by `wget http://www.dhondt.de/ccx_2.XX.src.tar.bz2` with XX the corresponding minor version of CalculiX and folder paths.
```

packaging/calculix-precice2_2.17-1_amd64/usr/share/doc/calculix-precice2/copyright
```diff
@@ -0,0 +1,13 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: preCICE
+Upstream-Contact: preCICE mailing-list <info@precice.org>
+Copyright: (c) 2019 the preCICE developers <info@precice.org>
```
MakisH, line 4: Question (I have no idea): Why 2019? Are we aiming for the oldest? The newest? A range? Is there a guideline?

packaging/README.md
MakisH, line 11: Very useful file!

The version is also hard-coded in the:
- `make_deb.sh`
- `DEBIAN/control`
- `changelog.Debian`
- `manpage.md`

We should comment on this here.

boris-martin: People never change :)

packaging/calculix-precice2_2.17-1_amd64/usr/share/doc/calculix-precice2/copyright
boris-martin: I copy pasted from the `libprecice` packages, so, no idea. I'm curious too. Guido Dhondt still uses a date from the last century, so probably not "newest".

MakisH: I understand that the year should be the date of publication (of what, I will touch in a moment). If we keep publishing, we should also make this a range of years. This is what I understand from this page: https://www.copyrightlaws.com/copyright-symbol-notice-year/

Now, the 2019 is probably the year we first had Debian packages for preCICE. If a package and the corresponding source are two different things is a different story. I would tend to write "2021" here, pointing to the package date. Or we add something like 2016-2021, covering the first state of the adapter, till today.

@uekerman do you maybe know more about this topic?

packaging/manpage.md
fsimonis: The official man page from man-pages on the convention for writing linux man pages :exploding_head: highly suggests to omit the AUTHOR section. https://linux.die.net/man/7/man-pages

packaging/calculix-precice2_2.17-1_amd64/usr/share/doc/calculix-precice2/copyright
fsimonis: @boris-martin You copy-pasted the output instead of the input :wink: 

https://github.com/precice/precice/blob/23ef51b8f429956ea620d8a9164e513bb074cd3a/tools/releasing/bumpversion.sh#L58-L64

boris-martin: Ah yes, didn't spot this automation, I could add it here.
Which means we'll end up with 2021 then ?
(I wonder how I got 2019 then, that's some old output)

fsimonis: You may :wink:

boris-martin: Actually I'm a bit confused (and very bad at reading Bash),  but this is more for the control file than copyright no?
(In the meantime I just updated to 2021 in the copyright)

I feel moderately comfortable adding this while we don't have a consistent versioning scheme.

MakisH: I think this is the information that package managers will show as license.
By the way, I notice there is a `Files: *` next to each license. Can we maybe use this to distinguish between CalculiX files and own files? Some files will have shared copyright, some will be distinct preCICE copyright.

boris-martin: That's the theory, but here the package only consists of the compiled executable, so it's not really useful. But I'd do this if I packaged the source so we can give single copyright to our helper files etc.

MakisH: Then everything is fine and I think we can resolve this discussion.

Makefile
MakisH: Do we want to rebase on develop first?

packaging/manpage.md
```diff
@@ -0,0 +1,35 @@
...
+
+: Prints current version of CalculiX. 
+
+**-i < jobname >**
+
+: Starts the job defined in <jobname>.inp. Be sure to omit the extension .inp in the job name.
+
+**-precice-participant < participant >**
+
+: Runs a preCICE simulation with name < participant >. This name must be the one used in the YAML config file as well as in the preCICE configuration file.
```
MakisH, line 25:
```suggestion
: Runs a preCICE simulation with name < participant >. This name must be the one used in the adapter configuration file as well as in the preCICE configuration file.
```

Makefile
boris-martin: I think yes. Otherwise the duplication of the Makefile would stay for a while.
(Then I don't know if I should change the target branch of the PR, or open a new one on the rebased after fixing locally the conflicts ?)

boris-martin:
@MakisH @KyleDavisSA I don't have the rights to fix the conflict, but it's a trivial one.
After this I think it's ready to merge

Makefile
MakisH: Done, there is no `Makefile` here anymore.

packaging/calculix-precice2_2.17-1_amd64/DEBIAN/control
```diff
@@ -0,0 +1,14 @@
+Package: calculix-precice2
+Version: 2.17-1
+Source: calculix-precice2
+Architecture: amd64
+Section: contrib/science
+Priority: optional
+Maintainer: The preCICE team (precice.org) <info@precice.org>
+Depends: libarpack2 (>= 2.1), libblas3 | libblas.so.3, libc6 (>= 2.29), libgcc-s1 (>= 4.0), libgfortran5 (>= 8), libgomp1 (>= 4.9), liblapack3 | liblapack.so.3, libprecice2 (>= 2.0.0), libspooles2.2, libstdc++6 (>= 9), libyaml-cpp0.6 (>= 0.6.2)
+Bugs: https://github.com/precice/calculix-adapter/issues
+Description: The official preCICE adapter for CalculiX.
```
MakisH, line 10: Lintian still complains with `W: calculix-precice2: description-synopsis-starts-with-article`. Maybe just write "

```suggestion
Description: Official preCICE adapter for CalculiX.
```

MakisH (approved):
- Fixed the conflict.
- I tried downloading the packages for 20.04 and doing some simple checks. The `sha256sum` seems to be correct. Lintian is happy apart from the one change suggested here.
- I have not tried installing the generated packages. Maybe try to install preCICE and the adapter from Debian packages in a clean Ubuntu 20.04 VM before merging?
- The rest looks good!

boris-martin:
It works on my working Ubuntu VM, worked on my Mint VM before this VM died, and on a clean WSL Ubuntu 20.04. I think tests are conclusive :)

MakisH (approved):

uekerman:
Nice! Please don't forget to also add this to the adapter documentation.

MakisH:
> Nice! Please don't forget to also add this to the adapter documentation.

Indeed, but we should probably start making releases first.

PR #73, opened 2021-11-10
Rebase v2.17 branch on develop

MakisH:
The adapter branch for CalculiX [v2.17](https://github.com/precice/calculix-adapter/commits/v2.17) was based on an older commit than the one provided by master/develop at that time. Combined with the fact that we are replacing complete CalculiX files every time in one commit, it was complicated to understand if we can actually rebase on `develop`, to get all the other changes that happened in the meanwhile. :spaghetti:

The main complication was the `nonlingeo_precice.c` file. After looking at the history together with @boris-martin, we decided that no not-formatting change happened in that file in the meanwhile from other commits. There was mainly [one commit](https://github.com/precice/calculix-adapter/commit/5b230dba339c55c67e6f30c0bd397d7a857b2b5a#diff-64bedaa3bfc160e6a2b54a758a76dec3a5d9a4fe5e5b49bccc374c02340f7675) by @IshaanDesai (moving the checkpoint writing), which [was later reverted](https://github.com/precice/calculix-adapter/commit/63af15333c8d5f3ba85fdec28ec35042cb2691e2#diff-64bedaa3bfc160e6a2b54a758a76dec3a5d9a4fe5e5b49bccc374c02340f7675).

If @KyleDavisSA agrees with the state of this branch, I can force-push to `v2.17` and delete this branch.

I think it would also be good to start using releases instead of version-specific branches, it could help prevent such issues in the future.


KyleDavisSA:
Is the aim to merge into develop or 2.17?

boris-martin:
> Is the aim to merge into develop or 2.17?

If I remember right, it's for develop, but at the same time the idea is that develop should be using 2.17 instead of 2.16

KyleDavisSA (approved):

MakisH:
> Is the aim to merge into develop or 2.17?

@KyleDavisSA Overall, we should use the same workflow as in the rest of the repositories:
- `master` for the releases
- `develop` accepting all PRs and for our testing

I would go even one step further and replace the version-specific branches with GitHub releases/tags, as we don't intend to keep maintaining older branches. This should also prevent such spaghetti situations in the future. We should only use version-specific branches when we develop for a new version, and then delete the branch.

So, I will force-push to `v2.17` now, and we decide on tags/releases/branches later.

MakisH:
I force-pushed the `v2.17_rebased_on_develop` to `v2.17`, so there is no more spaghetti. We can now continue with the rest of the PRs.

PR #75, opened 2021-11-25
Add a license compatibility statement

fsimonis:
Following the discussion in #72, we have to document our findings regarding the licensing somewhere.

Related to #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,15 @@ Go to the [adapter documentation](https://precice.org/adapter-calculix-overview.
 [1] Lucia Cheung Yau. Conjugate heat transfer with the multiphysics coupling library precice. Master’s thesis, Department of Informatics, Technical University of Munich, 2016.
 
 [2] Benjamin Uekermann, Hans-Joachim Bungartz, Lucia Cheung Yau, Gerasimos Chourdakis and Alexander Rusch. Official preCICE Adapters for Standard Open-Source Solvers. In Proceedings of the _7th GACM Colloquium on Computational Mechanics for Young Scientists from Academia_, 2017. 
+
+## License compatibility
+
+The calculix-adapter is licensed under the [GPLv3](./LICENSE).
+This is compatible with the GPLv2 License Calculix due to following statement:
+
+From the official [license published on dhondt.de](http://www.dhondt.de/gpl-2.0.txt)
+
+>    This program is free software; you can redistribute it and/or modify
+>    it under the terms of the GNU General Public License as published by
+>    the Free Software Foundation; either version 2 of the License, or
+>    (at your option) any later version.
--- a/README.md
+++ b/README.md
@@ -20,14 +20,6 @@ Go to the [adapter documentation](https://precice.org/adapter-calculix-overview.
 
 [2] Benjamin Uekermann, Hans-Joachim Bungartz, Lucia Cheung Yau, Gerasimos Chourdakis and Alexander Rusch. Official preCICE Adapters for Standard Open-Source Solvers. In Proceedings of the _7th GACM Colloquium on Computational Mechanics for Young Scientists from Academia_, 2017. 
 
-## License compatibility
+## License
 
-The calculix-adapter is licensed under the [GPLv3](./LICENSE).
-This is compatible with the GPLv2 License Calculix due to following statement:
-
-From the official [license published on dhondt.de](http://www.dhondt.de/gpl-2.0.txt)
-
->    This program is free software; you can redistribute it and/or modify
->    it under the terms of the GNU General Public License as published by
->    the Free Software Foundation; either version 2 of the License, or
->    (at your option) any later version.
+This project contains modified CalculiX files (licensed under [GPLv2 or later](http://www.dhondt.de/gpl-2.0.txt)) and additional adapter-specific files ([GPLv3](./LICENSE)). As a whole, this project is licensed under GPLv3.


MakisH:
I really like documentation, but isn't such a statement a bit unusual? Is this inspired by an example of some other project?

I am not sure how to express my concern, but it feels like this "is ok and should just work", therefore maybe the README is not the right place to add it, or maybe there is no need to mention it at all.

It definitely would prevent us from starting this discussion again, but maybe just keeping the discussion on GitHub is enough.

fsimonis:
Unusual yes, but so is patching a software and publishing it under a different license.

I would definitely add it somewhere prominent as GPLv2 and GPLv3 are incompatible. It just happens to be possible due to a changed statement in the file.

README.md
```diff
@@ -19,3 +19,15 @@ Go to the [adapter documentation](https://precice.org/adapter-calculix-overview.
...
+
+The calculix-adapter is licensed under the [GPLv3](./LICENSE).
+This is compatible with the GPLv2 License Calculix due to following statement:
+
+From the official [license published on dhondt.de](http://www.dhondt.de/gpl-2.0.txt)
+
+>    This program is free software; you can redistribute it and/or modify
+>    it under the terms of the GNU General Public License as published by
+>    the Free Software Foundation; either version 2 of the License, or
+>    (at your option) any later version.
```
MakisH, line 33: > I would definitely add it somewhere prominent as GPLv2 and GPLv3 are incompatible. It just happens to be possible due to a changed statement in the file.

I think this statement is not so uncommon and there is no problem at the moment. Now that we are happy, the only other stakeholders that could complain would be CalculiX (that already defined the license) or a third-party lawyer (who would immediately see the statement).

I am not against this change, I would just like the `README.md` to be short and useful. We could modify this to add more value, clarifying the two licenses of the repository. For example:

```suggestion
## License

This project contains modified CalculiX files (licensed under [GPLv2 or later](http://www.dhondt.de/gpl-2.0.txt)) and additional adapter-specific files ([GPLv3](./LICENSE)). As a whole, this project is licensed under GPLv3.
```

boris-martin:
> Unusual yes, but so is patching a software and publishing it under a different license.
> 
> I would definitely add it somewhere prominent as GPLv2 and GPLv3 are incompatible. It just happens to be possible due to a changed statement in the file.

I think it's more common than you think. The GPL 2/3 issue seems to have been crippling the free software world for a while and if I remember right the FSF recommends that everyone using GPL2 adds a compatibility statement.

MakisH:
> I think it's more common than you think. The GPL 2/3 issue seems to have been crippling the free software world for a while and if I remember right the FSF recommends that everyone using GPL2 adds a compatibility statement.

Do you have any reference for this FSF recommendation? I am mainly looking for examples on how it is common to implement this.

What if we add the compatibility statement directly in the license file? This is the place that someone interested to the license would look at. I just feel a bit uncomfortable adding a legal technicality of "all is well" to a user-facing document. Although we do want to keep the `README.md` only as a place for general information about the project.

> (BTW, shouldn't this change happen somewhere on the website too?)

I am not sure if that would be the right place. We can document the license there as a selling point, but I think this is a detail that belongs to the license file itself (or the README, with my considerations explained above).

boris-martin:
> > I think it's more common than you think. The GPL 2/3 issue seems to have been crippling the free software world for a while and if I remember right the FSF recommends that everyone using GPL2 adds a compatibility statement.
> 
> Do you have any reference for this FSF recommendation? I am mainly looking for examples on how it is common to implement this.

Well, simpler to quote you: 
> Regarding licensing, this is the license of CalculiX, as found in the `CalculiX.h` of version 2.17:
> 
> ```
> /*     CALCULIX - A 3-dimensional finite element program                 */
> /*              Copyright (C) 1998-2020 Guido Dhondt                     */
> 
> /*     This program is free software; you can redistribute it and/or     */
> /*     modify it under the terms of the GNU General Public License as    */
> /*     published by the Free Software Foundation; either version 2 of    */
> /*     the License, or (at your option) any later version.               */
> 
> /*     This program is distributed in the hope that it will be useful,   */
> /*     but WITHOUT ANY WARRANTY; without even the implied warranty of    */ 
> /*     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the      */
> /*     GNU General Public License for more details.                      */
> 
> /*     You should have received a copy of the GNU General Public License */
> /*     along with this program; if not, write to the Free Software       */
> /*     Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.         */
> ```
> 
> Since we have the `either version 2 of the License, or (at your option) any later version.`, I would say we have no problem with GPLv3.
> 
> From the [FSF article](https://www.gnu.org/licenses/license-compatibility.html) that @fsimonis posted above:
> 
> > The FSF uses the approach of asking people to release programs under “GNU GPL version N or any later version.” This licensing is compatible with version N, and also with N+1 (because it offers version N+1 as an option). When you combine code under “GPL 3 or later” with code under “GPL 2 or later”, the license of the combination is their intersection, which is “GPL 3 or later”.
> 
> The correct license of `calculix-precice`, as a code derived from CalculiX, and given our additions as GPLv3, is GPLv3.
> 
> In our case, we maintain the GPLv2 notice in `CalculiX.h`, but also provide GPLv3 for our code. I think this is ok but again, I barely understand licensing.

MakisH:
I am not sure I understand your latest comment.

CalculiX already has the `; either version 2 of the License, or (at your option) any later version.` statement in their `LICENSE`. We maintain this statement anyway.

But this PR proposes to add an additional explanation on our side, which confuses me.

I am mainly looking for an example of other projects adding such an additional explanation, I thought this is what you meant with the "not uncommon". This FSF quote does not mention something about an additional explanation in derived projects, it only explains what the result of the combination is.

I don't have any strong objections, so let's just add a short version of this in the `README`.

boris-martin:
Ah yes, I see. I just wanted to say that most it was recommended that all GPL2 projects should add a compatibility statement, but that's indeed not useful for us. Current proposal sounds fine to me

README.md
MakisH: @boris-martin @fsimonis any comments on this?

boris-martin: That's much shorter, I like it.

fsimonis:
So, what is the verdict now? :see_no_evil:

boris-martin:
> So, what is the verdict now? 🙈

This ? https://github.com/precice/calculix-adapter/pull/75#discussion_r757697634

Issue #10, opened 2018-10-31
Check for memory leaks


MakisH:
During tests for FSI with CalculiX and OpenFOAM, @derekrisseeuw noticed unreasonably high memory usage for CalculiX after a while, so there may be some memory leak in the adapter (or maybe in CalculiX itself).

At first, I would manually check if we delete every pointer we create. If this does not help, I would try to use Valgrind or a similar tool, although I would expect this to be a bit tricky in our case.

augustinhugues:
I am experiencing the same issue.
My calculation is crashing at some point with no apparent reason. If I look at the RAM usage it looks like it is increasing until reaching the maximum which might be the reason for the crash.
I am trying right now to run the same case with Calculix Only to check wether it is CalculiX or the adapter which is consumming memory.
I have 126GB of RAM, and at the beginning of the calculation, I clear cache, it it uses only 10GB for the first time steps.

augustinhugues:
[valgind-out.txt](https://github.com/precice/calculix-adapter/files/3266046/valgind-out.txt)

I use Valgind on ccx and got the folowwing file.

nkr1729:
@augustinhugues can you check if #16 clears this?

augustinhugues:
Hi @nkr1729 ,
I ran a simulation yesterday, it has been running for 18h hours now, and I don't have any increase of RAM consumption.

augustinhugues:
@nkr1729 
My bad, there was a problem in the calculation, CalculiX was not running, I re run and keep you update

nkr1729:
Thanks, can you also check with valgrind?

augustinhugues:
I checked with valgrind, there is only 47000bytes in use at exit against 5e6 so it is way better.
On my case the memory usage is still increasing but it's hard to say if the increase speed is reduced as it takes abour 90h to crash.

[valgrind-out_break.txt](https://github.com/precice/calculix-adapter/files/3319856/valgrind-out_break.txt)

nkr1729:
@augustinhugues thank you for checking. The current output looks like the leaks are not from calculix (or preCICE). I've no idea why, but, every leak in the report says `(in /bin/bash)`. Usually it should point to some file that is part of calculix or preCICE. And if the right flags are activated even a trace to the function from which the leak is coming. In your previous logs, we can see the trace going to `nonlingeo_precice.c` and `ccx_2.15.c`. However, there is nothing like that in this log.

xl305053:
[...]

PR #76, opened 2021-12-11
Fixed missing "free"

boris-martin:
While looking for some memory leaks ( https://github.com/precice/calculix-adapter/issues/10 ) I stumbled on this overlook.
I'm looking for bigger fish than this, but that's not a reason not to fix it.

There is a nested malloc in `Precice_Setup`, where we need to allocate an array of `adapterConfig.numInterfaces` different interfaces. So we create an array of pointers to interfaces, then populate it.
But in the cleanup function, we delete all the interfaces, but not the array itself.

Initialization : 
https://github.com/precice/calculix-adapter/blob/9fefcef8ade330280cb300c25c78df6827b44684/adapter/PreciceInterface.c#L39-L47

Cleanup : 
https://github.com/precice/calculix-adapter/blob/9fefcef8ade330280cb300c25c78df6827b44684/adapter/PreciceInterface.c#L417-L425

So I just added the required "free" call.
8 bytes less are leaked, a significant improvement :)

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -420,6 +420,7 @@ void Precice_FreeData( SimulationData * sim )
 		if( sim->preciceInterfaces[i] != NULL ){
 			free( sim->preciceInterfaces[i] );
 		}
+		free( sim->preciceInterfaces );
 	}
 
 	precicec_finalize();
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -418,9 +418,9 @@ void Precice_FreeData( SimulationData * sim )
 	{
 		PreciceInterface_FreeData( sim->preciceInterfaces[i] );
 		free( sim->preciceInterfaces[i] );
-		free( sim->preciceInterfaces );
 	}
-
+	
+	free( sim->preciceInterfaces );
 	precicec_finalize();
 }
 
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -417,9 +417,7 @@ void Precice_FreeData( SimulationData * sim )
 	for( i = 0 ; i < sim->numPreciceInterfaces ; i++ )
 	{
 		PreciceInterface_FreeData( sim->preciceInterfaces[i] );
-		if( sim->preciceInterfaces[i] != NULL ){
-			free( sim->preciceInterfaces[i] );
-		}
+		free( sim->preciceInterfaces[i] );
 		free( sim->preciceInterfaces );
 	}
 


KyleDavisSA:
Change seems reasonable. No problems with running a small test case to completion?

MakisH (approved):
If I understand correctly, `sim` is a structure provided to this setup function:

```c
void Precice_Setup( char * configFilename, char * participantName, SimulationData * sim )
{
```

and `sim->preciceInterfaces` are allocated in this function, as a pointer to a pointer:

```c
// Create interfaces as specified in the config file
sim->preciceInterfaces = (struct PreciceInterface**) calloc( adapterConfig.numInterfaces, sizeof( PreciceInterface* ) );

int i;
for( i = 0 ; i < adapterConfig.numInterfaces; i++ )
{
   InterfaceConfig * config = adapterConfig.interfaces + i;
	sim->preciceInterfaces[i] = malloc( sizeof( PreciceInterface ) );
	PreciceInterface_Create( sim->preciceInterfaces[i], sim, config );
}
```

I am very rusty in C (and needed to [check StackOverflow for common examples](https://stackoverflow.com/questions/5350314/how-to-free-a-pointer-to-a-dynamic-array-in-c), but I think it is safe to assume that, since we have two allocations, we should also have two deallocations.

This PR adds the `free` of the `*` next to the one for the `**` in the `void Precice_FreeData( SimulationData * sim )` teardown function.

Sounds reasonable to me, but please test this first.

boris-martin:
I already did the test, it runs exactly as before but with one less leaked pointer. Should I publish the output explicitly ?

MakisH:
It would always be nice to describe the methodology and document the output, but I would leave this up to you. :-)

adapter/PreciceInterface.c
```diff
@@ -420,6 +420,7 @@ void Precice_FreeData( SimulationData * sim )
 		if( sim->preciceInterfaces[i] != NULL ){
 			free( sim->preciceInterfaces[i] );
 		}
+		free( sim->preciceInterfaces );
```
fsimonis, line 423: This should be outside of the loop :wink:

adapter/PreciceInterface.c
```diff
@@ -420,6 +420,7 @@ void Precice_FreeData( SimulationData * sim )
 		if( sim->preciceInterfaces[i] != NULL ){
 			free( sim->preciceInterfaces[i] );
 		}
```
fsimonis, line 422: As you are already cleaning this up:

```suggestion
		free( sim->preciceInterfaces[i] );
```

`free(NULL);` does nothing by definition.

adapter/PreciceInterface.c
MakisH: oh, indeed, there is a `for` loop here! :see_no_evil:

boris-martin: Wooops.
Surprised it didn't blow then. Is that undefined behavior when you free the same thing multiple times ?

boris-martin: Hum now that I think about it, I tested locally then copy-pasted on Github instead of a  true commit from source, so it's probably just a typo. I'll run again to be sure.

boris-martin:
So, I re-ran again with the fix, and it's working: 
Valgrind finds exactly 8 less lost bytes, and the rest goes fine.

fsimonis:
I did the honours of merging this.

PR #78, opened 2022-01-23
Update to CalculiX 2.19

boris-martin:
CalculiX 2.18 was released a few months ago, but 2.19 came before we updated the adapter, so I wrote an updated version for it while skipping 2.18.

## Methodology

I took the files `nonlingeo.c`, `CalculiX` and `ccx_2.19.c` from CalculiX source code and reproduced changes by comparing 2.16 version of these files and 2.16 version of the adapter. Quite easy since most changes are documented or easy to spot (` /* Adapter */` comment or preCICE calls). I checked with `diff` (between CalculiX 2.16 and adapter 2.16) that I didn't forget any part. (By the way, I wish I had found earlier the option `diff -w` to ignore identation when comparing files!)
Then, simply updating the makefile did the trick.

## Status
- [x] Compiles successfully
- [x] Gives correct (good-looking) result on the "Partitioned Beam tutorial" ~~but crashes at "iteration 1 of time window 51 of 50". This doesn't seem to affect the output. Investigation coming, but if you spot it let me know!~~ **FIXED**, all works
- [x] Test on other tutorials

I don't know if there is something else to check (i.e. other tests than "running tutorials"), so let me know if something else is necessary.

## Miscellaneous

- I tried to be minimally invasive so I didn't change the indentation of the code, but it is actually terrible, so I wonder if I should format it ? This stuff is seriously hurting my eyes.
- For maintainability it would probably be useful to documente somewhere the steps for porting. If I remember right there was discussion with Guido Dhondt about integrating upstream, but I don't know the details nor how many times we'll have to play this little game again.


boris-martin:
Fixed the crash, I had forgotten a few stars when dealing with some pointers. 🙈 
I'll run a few other tutorials but I think it's ready-to-merge otherwise.

Makefile
```diff
@@ -69,7 +69,7 @@ FC = mpifort
 
 # Include a list of all the source files
 include $(CCX)/Makefile.inc
-SCCXMAIN = ccx_2.16.c
+SCCXMAIN = ccx_2.19.c
```
MakisH, line 72: There are a few more `2.16` further down (which actually don't matter, the number is not connected to the CalculiX version). Maybe it's a good chance to make this a variable?

MakisH (changes requested):
One thing to fix first in the Makefile.

Other than that, I tried running:

- perpendicular-flap: runs and results look reasonable.
- heat-exchanger: runs, but OpenFOAM explodes already at `Time = 5`. [Last time I checked](https://github.com/precice/precice/pull/1095#issuecomment-920164413), it was running fine (and I am using again preCICE v2.3.0, as released). I will investigate.
   - Update: Same for CalculiX 2.16.

Makefile
boris-martin: Probably good time for a variable yes.
Should we investigate the tutorial in the near future? (post release)

MakisH: I am investigating the tutorial at the moment, but this is independent of this PR.

MakisH: Related issue: https://github.com/precice/tutorials/issues/260

boris-martin:
Fixed the bad versioning and added a CCX_VERSION variable.
By the way, I added the flag for using OpenMP with Spooles (see https://github.com/precice/calculix-adapter/pull/34#issuecomment-1032077168 ) since my conclusion is that it should be there by default on makefiles of all branches, but this wasn't exactly the cleanest place to put it.

MakisH (approved):
Nice! The code now builds correctly and the perpendicular-flap is running. I have not looked at any code changes, as the diff is quite large (as usual).

KyleDavisSA:
I know it is a little late, but I did not find anything unusual about the file changes, so also happy its merged :).

Issue #87, opened 2022-02-20
Difference between faces-mesh and mesh ?


boris-martin:
I'm slightly confused by the different kinds of interfaces available.
From what I understand, we propose three different things:
- Nodal mesh without connectivity, for e.g. FSI with nearest neighbor or RBF mapping. Configuration is `nodes-mesh`
- Nodal mesh with connectivity, for same context but when nearest projection is used. Configuration is `nodes-mesh-with-connectivity`
- Face centers mesh for CHT. Configuration is either `mesh` or `faces-mesh`.

In the config reader code `mesh` and `faces-mesh` seem to have a different treatment. I think it's just a matter of style, but shouldn't it be a "if(mesh) else if (faces-mesh) else" instead of this "if-else-if" ?
I'm working on some refactoring but I'd rather have confirmation they're equivalent to be sure.

https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/ConfigReader.cpp#L36-L55

Also, documentation on [the website](https://precice.org/adapter-calculix-config.html) seems quite lacking. I'll work on it once I'm sure I don't have a major misunderstanding :)

MatthiasFreimuth:
I also use the faces-mesh for a FSI scenario in case of exchange-data pressure. So at least the statement that faces-mesh is for CHT only needs to be questioned in that case. (I added the pressure exchange data and will make a pull request with those changes soon)

boris-martin:
Glad to learn this! 
When I wrote faces were CHT-only, I meant it as "in the current state of the adapter" but pressure on faces definitely makes sense. Looking forward to the PR!

Issue #64, opened 2021-06-13
Requiring "Velocity" instead of "Velocities" breaks other adapters and schemes


j8asic:
At the line: https://github.com/precice/calculix-adapter/blob/d4db584237f5efdeb1f8da3d294d9766716ef711/adapter/PreciceInterface.c#L780
it says the requirement is that the tag must start with `"Velocity"`. Formerly the requirements was `"Velocities"`. Since we're talking about nodes, some prefer plural. Can you make the requirement to be: `if ( startsWith( config->writeDataNames[i], "Velocit" ) )` so both cases are covered?

IshaanDesai:
Thanks for this! It is fixed in https://github.com/precice/calculix-adapter/commit/caf3fe8004c6534ff2d5eeefcecaff2770a497f5

PR #88, opened 2022-02-20
Improved error message (for missing mesh)

boris-martin:
Current code contains a message to warn the user when he doesn't provide a nodal mesh (i.e. a face mesh is sent instead) despite being required. But the opposite wasn't true: if you send a nodal mesh when a face mesh is required, you get a very unhelpful "array out of bounds" crash, which was a pain to debug.

I thus added this error message. I also made them (both missing nodal and missing face mesh) slightly more explicit. Roughly speaking, instead of "You don't provide X", it is replaced by "you don't provide X, which is needed for using Y", with Y the cause of the error. (For instance if you try to read displacements on a face mesh, or heat flux on a nodal mesh).

**Examples**:
Writing a heat flux on a nodes mesh crashed without reason, now it says `Face centers mesh not provided in YAML config file. They are required for writing/reading the data Heat Flux.`

Reading a temperature on a face centers mesh used to give the message `Nodes mesh not provided in YAML config file` but now it gives `Nodes mesh not provided in YAML config file. They are required for writing/reading the data Temperature.`.

(I also fixed a typo where "Velocit" is used instead of "Velocity". This was fixed a while ago by @IshaanDesai but it seems we overrode it at some point when merging everything)

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -573,7 +573,7 @@ void PreciceInterface_NodeConnectivity(PreciceInterface *interface, SimulationDa
   PreciceInterface_ConfigureTetraFaces(interface, sim);
 }
 
-void PreciceInterface_EnsureValidNodesMeshID(PreciceInterface *interface, const char* type)
+void PreciceInterface_EnsureValidNodesMeshID(PreciceInterface *interface, const char *type)
 {
   if (interface->nodesMeshID < 0) {
     printf("Nodes mesh not provided in YAML config file. They are required for writing/reading the data %s.\n", type);
@@ -597,7 +597,8 @@ void PreciceInterface_ConfigureTetraFaces(PreciceInterface *interface, Simulatio
   }
 }
 
-void PreciceInterface_EnsureValidFacesMeshID(PreciceInterface* interface, const char *type) {
+void PreciceInterface_EnsureValidFacesMeshID(PreciceInterface *interface, const char *type)
+{
   if (interface->faceCentersMeshID < 0) {
     printf("Face centers mesh not provided in YAML config file. They are required for writing/reading the data %s.\n", type);
     fflush(stdout);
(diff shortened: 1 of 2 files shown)


adapter/PreciceInterface.c
```diff
@@ -670,44 +682,47 @@ void PreciceInterface_ConfigureCouplingData(PreciceInterface *interface, Simulat
...
     } else if (startsWith(config->writeDataNames[i], "Position")) {
-      PreciceInterface_EnsureValidNodesMeshID(interface);
+      PreciceInterface_EnsureValidNodesMeshID(interface, "Position");
       interface->writeData[i]    = POSITIONS;
       interface->positionsDataID = precicec_getDataID(config->writeDataNames[i], interface->nodesMeshID);
       printf("Write data '%s' found with ID # '%d'.\n", config->writeDataNames[i], interface->positionsDataID);
-    } else if (startsWith(config->writeDataNames[i], "Velocit")) {
-      PreciceInterface_EnsureValidNodesMeshID(interface);
+    } else if (startsWith(config->writeDataNames[i], "Velocity")) {
+      PreciceInterface_EnsureValidNodesMeshID(interface, "Velocity");
```
MakisH, line 720: I think this `Velocit` was actually the intended fix, to also work with `Velocities`. Since that already confused you, a comment would be useful here.

MakisH (approved):
I like this! It will probably help users that get confused at the mesh generation step.

How could I trigger this case, to test it?

adapter/PreciceInterface.c
```diff
@@ -573,10 +573,10 @@ void PreciceInterface_NodeConnectivity(PreciceInterface *interface, SimulationDa
   PreciceInterface_ConfigureTetraFaces(interface, sim);
 }
 
-void PreciceInterface_EnsureValidNodesMeshID(PreciceInterface *interface)
+void PreciceInterface_EnsureValidNodesMeshID(PreciceInterface *interface, const char *type)
 {
   if (interface->nodesMeshID < 0) {
-    printf("Nodes mesh not provided in YAML config file\n");
+    printf("Nodes mesh not provided in YAML config file. They are required for writing/reading the data %s.\n", type);
```
MakisH, line 579: I am wondering a bit about the action call: should the user change something in the adapter config file, or in the mesh itself? Maybe refering to the documentation would help.

adapter/PreciceInterface.c
boris-martin: Oooh right, I misunderstood the thing. For the record, explanation is here: https://github.com/precice/calculix-adapter/issues/64
I'm undoing this change.

adapter/PreciceInterface.c
boris-martin: Actually the original sentence is a bit misleading, as the problem is not *providing* the mesh, it's configuring correctly the YAML file. Given a mesh, either we configure the adapter to build a preCICE mesh from the nodes, or from the face centers. We can do both by configuring two interfaces in the YAML file (and creating two different preCICE meshes)

A different formulation would probably be more suited. Something like "This interface was configured to use nodes/faces, but it must be configured to use faces/nodes to transfer data Y". I'm open to suggestions.

To trigger the error, simply take any valid CalculiX test case and, in the `config.yaml` file, replace `faces-mesh` by `nodes-mesh` or vice-versa. (`mesh` seems to be an alias for `faces-mesh` and there's another one for connectivity, see https://github.com/precice/calculix-adapter/issues/87).
I first noticed this because it is actually impossible to do CHT with Dirichlet-Neumann coupling on the same mesh, since one requires nodes and the other one requires faces.

MakisH: I like your suggestion, let's go with that! :+1: Maybe even include some information about which "this interface" is?

Issue #90, opened 2022-02-23
Add support for MODAL DYNAMIC Modus of CalculiX


MatthiasFreimuth:
Information:

Target: Enable the use of MODAL DYNAMIC Cards within CalculiX in a preCICE-coupled simulation.

Implementation: done

Pull request: will hopefully follow soon

Issue #89, opened 2022-02-20
Add more supported elements to CHT simulations


boris-martin:
As of now, CHT simulations (more generally, simulations needing element face centers instead of nodes) only work with tetrahedrons.  (Probably *only first order tetrahedrons* but I haven't tested)
Using different elements (Say brick hexaedrons elements) crashes with no explicit messages but a segfault.

This part is problematic:

https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/PreciceInterface.c#L466-L472

- Function `PreciceInterface_ConfigureFaceCentersMesh` assumes tetrahedrons but should be adaptable +/- easily, as it calls a subroutine `getTetraFaceCenters` which is tetrahedron-specific
- PreciceInterface_ConfigureTetraFaces seems to be only required for connectivity but is always called (why?). This code is of course quite specific to the geometry.

I'd like to try to improve that in the near future, but I'm not 100% sure of the best approach:

- How do we detect element type ? Should we dive deeper into CalculiX code to detect each element and treat it accordingly ?
- Should we instead have only one type of elements (tetra or hexa) allowed (in the sense of "no mix") ? If so, is it detectable automatically, or should we add this as an option in the `config.yml` file?
- What about higher order elements ? wedge elements ? Supporting all elements could be complex.
- For connectivity, I read that supporte for quad elements on the interface is experimental. Is it safe to use ?

MatthiasFreimuth:
Hi Boris,
I actually already implemented some way of support for hexahedral elements, in my case I needed it for a DLOAD card, so to apply pressure values in CalculiX. I'm about to make a pull request with my changes soon and then we can use this as a starting point for further discussions about this topic.

boris-martin:
Looking forward to it, thanks! It can probably wait for the actual PR, but how do you handle mixed meshes ? Do you detect for each element what kind it is? Do you only detect the element type of a single-type mesh? Or is that a configuration thing ?

MatthiasFreimuth:
Actually I only use single-type meshes, so no mixed meshes. One can detect the element type directly from the CalculiX variable lakon. You can find it in the list of variables in the CalculiX documentation.

PR #91, opened 2022-03-02
Added Pressure ExchangeData, hexaedral elements and Modal Dynamic Modus

MatthiasFreimuth:
This PR enhances the adapter with respect to the following functionalities:

- Support of Pressure as a read exchange variable
- Support of faces for hexaedral elements e.g. to create facecenters meshes using hexaedral elements (without connectivity)
- Support of the MODAL_DYNAMIC Modus of CalculiX

Closes issue #90 Related to issue #89 (Related to issue #87)

Once this PR is accepted the new functionalities should be described in the adapter's documentation.

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -216,6 +216,11 @@ void Precice_ReadCouplingData(SimulationData *sim)
           setNodeForces(interfaces[i]->nodeVectorData, interfaces[i]->numNodes, interfaces[i]->dimCCX, interfaces[i]->xforcIndices, sim->xforc);
           printf("Reading FORCES coupling data with ID '%d'. \n", interfaces[i]->forcesDataID);
           break;
+        case PRESSURE:
+          precicec_readBlockScalarData(interfaces[i]->pressureDataID, interfaces[i]->numElements, interfaces[i]->preciceFaceCenterIDs, interfaces[i]->faceCenterData);
+          setFacePressure(interfaces[i]->faceCenterData, interfaces[i]->numElements, interfaces[i]->xloadIndices, sim->xload);
+          printf("Reading PRESSURE coupling data with ID '%d'. \n", interfaces[i]->pressureDataID);
+          break;
         case DISPLACEMENTS:
           // Read and set displacements as single point constraints (Dirichlet BC)
           if (isQuasi2D3D(interfaces[i]->quasi2D3D)) {
@@ -280,7 +285,7 @@ void Precice_WriteCouplingData(SimulationData *sim)
                                     sim->istartset,
                                     sim->iendset,
                                     sim->ipkon,
-                                    *sim->lakon,
+                                    sim->lakon,
                                     sim->kon,
                                     sim->ialset,
                                     sim->ielmat,
@@ -317,7 +322,7 @@ void Precice_WriteCouplingData(SimulationData *sim)
                             sim->istartset,
                             sim->iendset,
                             sim->ipkon,
-                            *sim->lakon,
+                            sim->lakon,
                             sim->kon,
                             sim->ialset,
                             sim->ielmat,
@@ -435,6 +440,7 @@ void PreciceInterface_Create(PreciceInterface *interface, SimulationData *sim, I
   interface->positionsDataID              = -1;
   interface->velocitiesDataID             = -1;
   interface->forcesDataID                 = -1;
+  interface->pressureDataID               = -1;
 
   // Check if quasi 2D-3D coupling needs to be implemented
   if (interface->dim == 2) {
@@ -487,7 +493,13 @@ void PreciceInterface_ConfigureFaceCentersMesh(PreciceInterface *interface, Simu
 
   interface->faceCenterCoordinates = malloc(interface->numElements * 3 * sizeof(double));
   interface->preciceFaceCenterIDs  = malloc(interface->numElements * 3 * sizeof(int));
-  getTetraFaceCenters(interface->elementIDs, interface->faceIDs, interface->numElements, sim->kon, sim->ipkon, sim->co, interface->faceCenterCoordinates);
+  if (startsWith(&sim->lakon[1*8], "C3D4") || startsWith(&sim->lakon[1*8], "C3D10")){
+    getTetraFaceCenters(interface->elementIDs, interface->faceIDs, interface->numElements, sim->kon, sim->ipkon, sim->co, interface->faceCenterCoordinates);
+  } else if (startsWith(&sim->lakon[1*8], "C3D8") || startsWith(&sim->lakon[1*8], "C3D20")){
+    getHexaFaceCenters(interface->elementIDs, interface->faceIDs, interface->numElements, sim->kon, sim->ipkon, sim->co, interface->faceCenterCoordinates);
+  } else{
+    supportedElementError();
+  }
 
   interface->faceCentersMeshID    = precicec_getMeshID(interface->faceCentersMeshName);
   interface->preciceFaceCenterIDs = malloc(interface->numElements * sizeof(int));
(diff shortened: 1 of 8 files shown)


adapter/PreciceInterface.c
```diff
@@ -487,7 +493,13 @@ void PreciceInterface_ConfigureFaceCentersMesh(PreciceInterface *interface, Simu
 
   interface->faceCenterCoordinates = malloc(interface->numElements * 3 * sizeof(double));
   interface->preciceFaceCenterIDs  = malloc(interface->numElements * 3 * sizeof(int));
-  getTetraFaceCenters(interface->elementIDs, interface->faceIDs, interface->numElements, sim->kon, sim->ipkon, sim->co, interface->faceCenterCoordinates);
+  if (startsWith(&sim->lakon[1*8], "C3D4") || startsWith(&sim->lakon[1*8], "C3D10")){
```
boris-martin, line 496: Why read at `1*8` instead of 0?

ccx_2.19.c
```diff
@@ -1445,8 +1446,41 @@ int main(int argc, char *argv[])
           printf("ERROR: This simulation type is not available with preCICE");
           exit(0);
         }
-      }
-
+      } else if (isModalDynamic){
+
+        printf("Starting FSI analysis via preCICE\n");
```
boris-martin, line 1451: Some print to acknowledge we use modal dynamic seems appropriate.

dyna_precice.c
```diff
@@ -0,0 +1,2177 @@
...
+	  char *prset,ITG *nener,double *trab,
+	  ITG **inotrp,ITG *ntrans,double **fmpcp,ITG *ipobody,ITG *ibody,
+	  double *xbody,ITG *nbody,double *xbodyold,ITG *istep,
+	  ITG *isolver,ITG *jq,char *output,ITG *mcs,ITG *nkon,
+	  ITG *mpcend,ITG *ics,double *cs,ITG *ntie,char *tieset,
+	  ITG *idrct,ITG *jmax,
+	  double *ctrl,ITG *itpamp,double *tietol,ITG *nalset,
+	  ITG *ikforc,ITG *ilforc,double *thicke,ITG *nslavs,ITG *nmat,
+	  char *typeboun,ITG *ielprop,double *prop,char *orname,
+	  double *t0g,double *t1g
```
boris-martin, line 78:
```suggestion
	  double *t0g,double *t1g,
```

There's a missing comma that prevents the build.

boris-martin:
Hi @MatthiasFreimuth,

Thank you very much for this PR. I already had a quick look but will make a more complete review in a few days.
A few questions to be sure I understand everything:

- Pressure can be read, because we convert it to a distributed load in CalculiX. However there is no Pressure write, because Calculix only outputs forces on nodes, but no pressure. Right ?
- For MODAL DYNAMIC simulations, we must first compute the eigenmodes, then configure the simulation with this keyword, and everything else related to the adapter works the same way? Or are there necessary precautions? I read that `Temperature loading or residual stresses are not allowed`, but I'm not good enough with solid mechanics to fully understand this. Residual stress only occur when using non elastic material, if I understood it right.
- I see that for faces mesh, you look whether the first (second?) element is a tetrahedron or a hexaedron. Do you think it would be doable to refactor the code to allow mixed meshes? I'm thinking of some kind of loop over all elements where for each element we look up its type, but maybe there's a pitfall I didn't see. I imagine for now you implemented non mix mesh because that was enough for your requirements.

MatthiasFreimuth:
Hi Boris,
regarding your questions: 

- > Pressure can be read, because we convert it to a distributed load in CalculiX. However there is no Pressure write, because Calculix only outputs forces on nodes, but no pressure. Right ?
Actually I just implemented pressure as a read variable, because I need it that way and never thought about wether writing pressure is possible or would be useful for any kind of simulation. I'm not sure if CalculiX only outputs forces. 

- > For MODAL DYNAMIC simulations, we must first compute the eigenmodes, then configure the simulation with this keyword, and everything else related to the adapter works the same way? Or are there necessary precautions? I read that Temperature loading or residual stresses are not allowed, but I'm not good enough with solid mechanics to fully understand this. Residual stress only occur when using non elastic material, if I understood it right.
For MODAL DYNAMIC simulations you first need to calculate the eigenmodes, which can be stored by CalculiX in a file with the ending ".eig". (This doesn't have to be a coupled simulation!) I guess everything else related to the adapter works similar, even though I'm not quite sure wether the adapter is able to perform a clean exit of the simulation or not. Regarding your question about the temperature loading and residual stresses I'm unsure too, as my background isn't in structural mechanics either.

- > I see that for faces mesh, you look whether the first (second?) element is a tetrahedron or a hexaedron. Do you think it would be doable to refactor the code to allow mixed meshes? I'm thinking of some kind of loop over all elements where for each element we look up its type, but maybe there's a pitfall I didn't see. I imagine for now you implemented non mix mesh because that was enough for your requirements.
I did not think about mixed meshes yet, because i did not need them, thats true. But right now I see no pitfalls for the implementation for mixed meshes you suppose.

ccx_2.19.c
MatthiasFreimuth: Do you think this is really needed here? Thats something you configured in your CalculiX-.inp file and there is some output from CalculiX itself that it composes the dynamic response from the eigenmodes. We don't say that we use the dynamic modus either if we do so right? So on my opinion the information that belongs to preCICE is that we are doing an FSI analysis but i'm not sure if mentioning the method of the solid solver is appropriate here.

boris-martin: Ah yes, if CalculiX says so, it's probably enough.

adapter/PreciceInterface.c
MatthiasFreimuth: 0 should also work.

boris-martin:
I tried running the "perpendicular flap" tutorial with a modal dynamic analysis and it crashes (apparently because Quasi-Newton acceleration doesn't converge). Did you manage to get it running? Or am I using it wrong? I simply used the default .inp file and changed the DYNAMIC analysis with a FREQUENCY followed by a MODAL DYNAMIC analysis.

```
*INCLUDE, INPUT=all.msh
*INCLUDE, INPUT=fix1_beam.nam
*INCLUDE, INPUT=interface_beam.nam
*MATERIAL, Name=EL
*ELASTIC
 4000000, 0.3
*DENSITY
 3000
*SOLID SECTION, Elset=Eall, Material=EL
*STEP, INC=1000000

*BOUNDARY
Nfix1, 1, 3
*BOUNDARY
Nall, 3

*FREQUENCY, STORAGE=YES
780

*MODAL DYNAMIC DIRECT
1.E-2, 5.0

*CLOAD
 Nsurface, 1, 0.0
 Nsurface, 2, 0.0
 Nsurface, 3, 0.0
*NODE FILE
 U
*EL FILE
 S, E
*END STEP
```

MatthiasFreimuth:
Hi Boris,
as far as I know the Frequency card and the Modal dynamic card should each be a separate step. I always do two separate simulations, one to generate the .eig file with CalculiX (FREQUENCY; STORAGE=YES as STEP) and then the coupled simulation (MODAL DYNAMIC as STEP). The approach to leave the rest of the input file the same is basically correct. 
I did not try any of the tutorial cases with the new features yet, but for the testcases I used it worked fine, but there I only used  explicit coupling so far.

boris-martin:
Hi Matthias,

It converges now, having both frequency and modal dynamic in one step seems to be what broke the simulation.
However now I face the problem of a zero displacement of the flap (i.e. nothing happens)
I don't know yet whether the reading fails or somehow the object if fixed. Any tips ?

Frequency file:

```
*INCLUDE, INPUT=all.msh
*INCLUDE, INPUT=fix1_beam.nam
*INCLUDE, INPUT=interface_beam.nam
*MATERIAL, Name=EL
*ELASTIC
 4000000, 0.3
*DENSITY
 3000
*SOLID SECTION, Elset=Eall, Material=EL

*BOUNDARY
Nfix1, 1, 3
*BOUNDARY
Nall, 3

*STEP, INC=1000000
*FREQUENCY, STORAGE=YES
750
*END STEP
*NODE FILE
 U
*EL FILE
 S, E

*CLOAD
 Nsurface, 1, 0.0
 Nsurface, 2, 0.0
 Nsurface, 3, 0.0
*NODE FILE
 U
*EL FILE
 S, E
*END STEP
```

Dynamic file:

```
*INCLUDE, INPUT=all.msh
*INCLUDE, INPUT=fix1_beam.nam
*INCLUDE, INPUT=interface_beam.nam
*MATERIAL, Name=EL
*ELASTIC
 4000000, 0.3
*DENSITY
 3000
*SOLID SECTION, Elset=Eall, Material=EL

*BOUNDARY
Nfix1, 1, 3
*BOUNDARY
Nall, 3

*STEP, INC=1000000
*MODAL DYNAMIC DIRECT
1.E-2, 5.0

*CLOAD
 Nsurface, 1, 0.0
 Nsurface, 2, 0.0
 Nsurface, 3, 0.0
*NODE FILE
 U
*EL FILE
 S, E
*END STEP
```

Also, I noticed running the frequency analysis with preCICE doesn't work. (The adapter says that preCICE doesn't support frequency analysis)
So I need to run without preCICE participant to compute the frequency first. Shouldn't we tweak this to ismply discard preCICE if frequency analysis is selected?

adapter/PreciceInterface.c
KyleDavisSA: Is keeping it 1*8 fine?

boris-martin: I think as long as the mesh uses only one type of elements it should work. In the long run if we want to support "hybrid" meshes we should change this but for now I expect it to work. (Unless maybe there is exactly one element?)

KyleDavisSA:
The perpendicular flap moves when I use explicit coupling, but not implicit coupling. Examining `dyna_precice.c`, I see there is no variable 'icutb', which I think controls the time stepping if I am not mistaken. This variable is used to reset the displacements when convergence is not reached. I am happy to merge this while only supporting explicit coupling for now, and we can work on extending to implicit coupling.

Issue #86, opened 2022-02-15
Null essential (Dirichlet) boundary condition imposed on node 1


AndresPedemonteFIUBA:
Hey,

I have experienced a problem with the adapter for CalculiX 2.16 where node number 1 of the CalculiX mesh has a null Dirichlet BC imposed on it for every increment/time step of a simulation. This happens even when there are no BCs specified for that node. In the case of an FSI simulation this means node 1 stays fixed; it's displacement is always zero. 

Others have reported this issue as well:

- [CalculiX single node stagnation problem](https://precice.discourse.group/t/calculix-single-node-stagnation-problem/518) (Specially, check out post number 3).
- [FSI SU2/CalculiX (flap_perp): CalculiX node ID#1 is fixed](https://github.com/precice/tutorials/issues/90)
- [How can I create a Solid mesh for working with the nearest neighbor mapping?](https://precice.discourse.group/t/how-can-i-create-a-solid-mesh-for-working-with-the-nearest-neighbor-mapping/243) (See post number 4 by Kyle Davis)
- [Temperature anomaly/ Calculix OpenFoam CHT Coupling](https://precice.discourse.group/t/temperature-anomaly-calculix-openfoam-cht-coupling/915)

The last one seems to indicate this problem also happens for CHT simulations. In that case, for node number 1 T = 0 K.

An easy way to reproduce it is by modifying the `perpendicular-flap` tutorial. Just exchange nodes `1` and `738` in the `*NODE` and `*ELEMENT` keywords. Also, change node `1` to `738` in node set `Nfix1` (`fix1_beam.nam` file). You should get something like this:
![Screenshot from 2022-02-15 20-36-56](https://user-images.githubusercontent.com/67233283/154168477-eae766a0-142d-4ed2-ae6f-621770827cf4.png)
That fixed vertex is node `1`. This is for `t=0.3 s`; CalculiX (adapter for 2.16) coupled with OpenFOAM (v7 .org, version 1.0 of the adapter), preCICE 2.3.0, Ubuntu 20.04 LTS.

I have not tested the adapter for CalculiX 2.19 and cannot say wether this shows up with it.

Best regards,
Andrés

boris-martin:
Hi Andres. Thanks for finding out and documenting this. I successfully managed to reproduce it (the perpendicular flap) with both 2.16 and 2.19 versions of the adapter. I'll investigate this :)

[...]

PR #92, opened 2022-03-07
2D simulations major overhaul

boris-martin:
# Context

Currently, simulations in Quasi2D-3D are limited to meshes where there is exactly 2 points in 3D space for a 2D point: one with z=0 and one with z=1. In particular, this forbids the use of 2nd order elements. Furthermore, mapping from 3D to 2D is always consistent but mapping from 2D to 3D is conservative. For instance, reading forces and sending displacements works but the opposite is broken. Having a Temperature Dirichlet BC is not possible (it would be divided by 2)

# Approach

Mapping is reworked to allow for a variable number of 3D points per 2D point. Both consistent and conservative mappings are implemented. Number of 3D points on a 2D point is used to compute average in a consistent write, or to split loads in conservate reads.
Code is refactored to limit the number of 2D-specific code inside `PreciceInterface.c` (which is 100 lines shorter!)

# New features

- Allows higher order meshes in 2D
- Allows actual 2D elements to be used (plane stress / plane strain as well as shells)
- Should fix reading displacements & temperatures

# Checks

- [ ] Perpendicular flap with 2nd order elements
- [x] Perpendicular flap with 2D plane stress elements: ~~done, but results inconsistent. Investigation ongoing.~~ Shell elements give good results, unlike the plane stress elements. Apparently 1st order Shell expands into C38DI whereas plane stress expand into C3D, which probably explains it. Anyway, 2D coupling with 2D elements works, in terms of code
- [x] Non-regression test for current perpendicular flap
- [x] Thermal simulation to check Temperature reading
- [x] Check for dead code to remove or forgotten cleanup

# Future work

- Some refactoring could be done for 3D coupling
- This only works for node mesh. Making stuff work on face meshes could be interesting.


boris-martin:
Test case for temperature: https://github.com/boris-martin/ccx_heat_2d_test
It's a simple heat conduction in a cubic domain with one side at 300K and another side where Temperature is read from preCICE, with a simple script that always sends T = 350K.
It works in 3D or in 2D with this PR, but using this in 2D without the PR reads a temperature of 175K instead.

(On a side note, this is a case that spontanously reproduces issue #86, which I struggled to reproduce before. Life is weird that way)

IshaanDesai (approved):
This is a really significant improvement to the 2D-3D case handling capability and is a general and clean solution! Great work :+1: I did not look at each and every line of the change, but I ran the perpendicular-flap tutorial case with OpenFOAM and SU2 and the results are as expected.

uekerman:
Hi @boris-martin 
Wouldn't your testcase already (almost) be a solution to https://github.com/precice/tutorials/issues/103 ?

boris-martin:
@uekerman Exactly, I was trying to set up this case then realized all the missing features on the adapter and started this PR.
There are a few more things to work on. For instance 2D coupling works for nodal meshes but not meshes with face centers, required for heat fluxes. But this new tutorial case is a milestone I've set up :)
I'd also like to add the support for brick elements in CHT (only tetrahedra are supported yet) which is part of https://github.com/precice/calculix-adapter/pull/91 but wasn't yet adapted to 2D.................

PR #94, opened 2022-04-09
Fixed mesh configuration 2D coupling with Face meshes

boris-martin:
# Context

Quasi2D simulations are currently supported with nodal meshes (for FSI typically) but weren't supported for faces mesh. (i.e. to transfer pressure, heat fluxes, film temperature or convection coefficient, so mostly heat transfer)
While with nodal meshes we need some elaborate tricks to convert from 3D to 2D (because there is more than one 3D vertex for a 2D vertex), face meshes are simpler: just cropping the Z-component should work. Since all mappings are consistent, there is no need to pay attention to the number of faces etc.

# Changes

When configuring a face mesh in 2D, nothing changes except that if preCICE is configured in 2D, we crop the Z-component when sending the mesh. Previously, sent data was erroneous (CalculiX sends data in the format "x1 y1 z1 x2 y2 z2..." but preCICE expects 2D data so the x-coordinate of the second vertex is z1, and so on).

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -511,7 +511,6 @@ void PreciceInterface_ConfigureFaceCentersMesh(PreciceInterface *interface, Simu
   interface->preciceFaceCenterIDs = malloc(interface->numElements * sizeof(int));
 
   sendFaceCentersVertices(interface);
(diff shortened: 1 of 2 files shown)


IshaanDesai (approved):
The thought behind the change and the code look good, although I did not test it explicitly. This PR has been referenced for a test case with CalculiX and hence some confirmatory testing is done.

PR #95, opened 2022-05-16
Fixed crash when using face meshes and element IDs don't start to 0

boris-martin:
# Context

Since the merge of https://github.com/precice/calculix-adapter/pull/91, simulations using face meshes can use either tetrahedra elements or hexaedral elements, instead of just tetrahedra. (It is however still limited to cases with only one type)
Current implementation finds whether tetra or hex are used by looking at the type of the first element (since we assume they're all the same). However, CalculiX can use meshes where there is no first element, in the sense that indices can start after 0. (See mesh of the heat exchanger tutorial for instance)
When this occurs, we find a "null" element and return an error because of the use of an unsupported element.

# Proposed fix

Iterate over all elements until a non-null element is found, and use this one to decide between tetra and hexas.

# Reproduce the bug

Try to run heat exchanger tutorial with current `develop` branch: it outputs an error because of unsupported element.


adapter/CCXHelpers.h
```diff
@@ -53,6 +53,17 @@ enum CouplingDataType { TEMPERATURE,
                         VELOCITIES,
                         POSITIONS };
 
+/**
+ * @brief Type of element used for faces mesh, where we assume only one type of element is used.
+ *  TETRAHEDRA      - C3D4 or C3D10 element
+ *  HEXAEDRA        - C3D8 or C3D20 elment
```
IshaanDesai, line 59:
```suggestion
 *  HEXAHEDRA        - C3D8 or C3D20 elment
```

adapter/CCXHelpers.h
```diff
@@ -53,6 +53,17 @@ enum CouplingDataType { TEMPERATURE,
...
+ *  INVALID_ELEMENT - Anything else
+ */
+enum ElemType { TETRAHEDRA,
+                HEXAEDRA,
```
IshaanDesai, line 63:
```suggestion
                HEXAHEDRA,
```

adapter/PreciceInterface.c
```diff
@@ -486,9 +486,28 @@ void PreciceInterface_Create(PreciceInterface *interface, SimulationData *sim, I
...
+  // Assuming only tetrahedra are used, or only hexaedral, for faces meshes.
+  // Return first non-zero mesh type.
+  // lakon tab takes 8 chars per element
+
+  const char *lakon_ptr = sim->lakon;
+  for (int i = 0; i < sim->ne; ++i) {
+    if (startsWith(lakon_ptr, "C3D4") || startsWith(lakon_ptr, "C3D10")) {
+      return TETRAHEDRA;
+    } else if (startsWith(lakon_ptr, "C3D8") || startsWith(lakon_ptr, "C3D20")) {
+      return HEXAEDRA;
```
IshaanDesai, line 500:
```suggestion
      return HEXAHEDRA;
```

adapter/PreciceInterface.c
```diff
@@ -499,9 +518,14 @@ void PreciceInterface_ConfigureFaceCentersMesh(PreciceInterface *interface, Simu
...
   interface->preciceFaceCenterIDs  = malloc(interface->numElements * 3 * sizeof(int));
-  if (startsWith(&sim->lakon[1 * 8], "C3D4") || startsWith(&sim->lakon[1 * 8], "C3D10")) {
+
+  enum ElemType elemType = findSimulationMeshType(sim);
+
+  if (elemType == TETRAHEDRA) {
+    printf("Configuring faces mesh with tetrahedra.\n");
     getTetraFaceCenters(interface->elementIDs, interface->faceIDs, interface->numElements, sim->kon, sim->ipkon, sim->co, interface->faceCenterCoordinates);
-  } else if (startsWith(&sim->lakon[1 * 8], "C3D8") || startsWith(&sim->lakon[1 * 8], "C3D20")) {
+  } else if (elemType == HEXAEDRA) {
```
IshaanDesai, line 527:
```suggestion
  } else if (elemType == HEXAHEDRA) {
```

IshaanDesai (commented):
Had a brief look, did not try to run the heat-exchanger tutorial. Running that tutorial would be a good check for this right?

boris-martin:
@IshaanDesai thanks for the review!
Yes, running the heat exchanger is a good test and it worked for me. (it still suffers from the "crash at t=5" issue we recently reopened, but without this PR it crashes earlier at initalization time)

adapter/CCXHelpers.h
IshaanDesai: @boris-martin this change is still not included, and hence I cannot compile the adapter for now, to test the heat-exchanger case.

boris-martin: My bad! Done

PR #99, opened 2022-07-17
Fixed implicit coupling in modal dynamic simulation.

### LINKED ISSUE #93: Fix broken implicit coupling when using modal dynamic simulations.
Since https://github.com/precice/calculix-adapter/pull/91, the adapter supports modal dynamic simulations. (Where we use pre-computed eigenmode decomposition to speedup the computations by projecting forces into these modes and solving harmonic oscillators)
However, this functionality was designed with explicit coupling in mind and doesn't work with implicit coupling. It seems null displacements are generated. I added a warning message until we fix this.

boris-martin:
See https://github.com/precice/calculix-adapter/issues/93.
Implicit coupling now seems to work as expected, I have a modified version of the perpendicular flap tutorial adapted to modal dynamic simulations that gives plausible results (although slightly different than the reference solution, but since I used a smaller number of frequencies it makes sense).

The fix is quite simple to understand (but was tricky to find in the hundred of lines of badly-documented code :sweat_smile: ) : since the simulation is done in frequency space, checkpointing must restore DOFs in the corresponding eigenmodes, not just the final displacements. So when reading I checkpoint I update these.

I also removed the use of the `icutb` variable, which is actually unused

### DIFF
--- a/dyna_precice.c
+++ b/dyna_precice.c
@@ -1432,11 +1432,11 @@ void dyna_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **lakonp
     if (Precice_IsWriteCheckpointRequired()) {
       printf("WARNING: implicit coupling with modal dynamic simulations is not working in the current version of the adapter.\n");
       Precice_WriteIterationCheckpoint(&simulationData, vini);
+      // Otherwise, each iteration in implicit coupling would be written as a new step
+      iinc++;
+      jprint++;
       Precice_FulfilledWriteCheckpoint();
     }
-
-    iinc++;
-    jprint++;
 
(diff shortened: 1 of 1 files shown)


boris-martin:
Marking this as a draft for now, because there seems to be a bug that makes non-converging iterations written as output too.

boris-martin:
Fixed that part !
As a quick demo, comparison of the perpendicular flap with regular simulation (grey) VS modal dynamic with only 10 eigenmodes (wireframe):

flapmodal

That's much more accurate than I anticipated, so I call this a win :)

KyleDavisSA (approved):
All changes seem good!

Issue #63, opened 2021-05-18
Redesigning of CalculiX adapter which allows ''STATIC" step for FSI simulation


rahuldevmeher:
Hello,

I want to do a steady-state FSI simulation using CalculiX as solid solver. Currently the CalculiX adapter only allows a “DYNAMIC” step for FSI. Whereas for doing a steady-state FSI, I need to use a “STATIC” step in CalculiX. If I use the keyword “**STATIC**” in CalculiX input file, the following error is shown in Solid.log - 
`**ERROR: Only thermal coupling or FSI is available with preCICE**`

So, what changes need to be made in the adapter to implement this new feature where the adapter allows a “STATIC” step as well ? Suggestions are requested to implement this feature.

MatthiasFreimuth:
Hello,
I think the main changes would be:
1. Adapt the ccx_2.16.c file to enable further cases, where preCICE can be used here: https://github.com/precice/calculix-adapter/blob/9fefcef8ade330280cb300c25c78df6827b44684/ccx_2.16.c#L1109 
2. Adapt the linstatic.c routine to be able to interact with preCICE (include the preCICE-API in the main iteration/increment loop of the routine) Similar to a nonlingeo_precice.c there should then be a linstatic_precice.c
3. Include those changed files in the Makefiles accordingly
 
For my best knowledge that would be the general procedure to enable other simulation types of CalculiX for a preCICE simulation as well.

IshaanDesai:
@MatthiasFreimuth thanks for your input!
@rahuldevmeher if you want, you can start modifying the adapter and also simultaneously open a draft pull request so that I can also see the changes and help with testing. You can then also assign yourself to this issue which would indicate to other users and developers that you are developing this feature.

MatthiasFreimuth:
I just figured out there is a simpler way to enable Static steps for FSI simulations in the adapter. 
If one specifies the following in the CalculiX input file:

*STEP, NLGEOM
*STATIC
1,1
*END STEP

(so basically a STATIC step with the NLGEOM option, so that it is solved by the nonlingeo routine)
one does not need to adapt the linstatic.c routine and a small adaption of the ccx_2.16.c(or higher) file is enough.

boris-martin:
So if I understand correctly (using adapter version from your PR):
- If you want to do a non-linear static analysis, you configure a STEP, NLGEOM with STATIC then define loads
[...]

PR #101, opened 2022-07-21
Enable Static Step for FSI

MatthiasFreimuth:
Enable basic Static Step for FSI

One use case could be a one-way coupled static precalculation to take a deformation due to steady loads into account before the real instationary fully coupled simulation starts. One can therefore use the restart capabilities of CalculiX.

related to #63

### DIFF
--- a/ccx_2.19.c
+++ b/ccx_2.19.c
@@ -1312,6 +1312,7 @@ int main(int argc, char *argv[])
       int isDynamic         = ((nmethod == 4) && (iperturb[0] > 1));
       int isThermalAnalysis = ithermal[0] >= 2;
       int isModalDynamic    = ((nmethod == 4) && (iperturb[0] < 2));
+      int isStaticNLGEOM = ((nmethod==1) && (iperturb[0]>1));
 
       if (isStaticOrDynamic && isThermalAnalysis) {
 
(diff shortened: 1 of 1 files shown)


boris-martin:
I tried testing this by modifying the elastic-tube-3d tutorial.
I added a static step before the dynamic step:

```
*INCLUDE, INPUT=all.msh
*INCLUDE, INPUT=inlet.nam
*INCLUDE, INPUT=outlet.nam
*INCLUDE, INPUT=wall.nam
*INCLUDE, INPUT=interface.nam
*INCLUDE, INPUT=interface.sur
*MATERIAL, Name=EL
*ELASTIC
300000.0, 0.3
*DENSITY
1200.0
*SOLID SECTION, Elset=Eall, Material=EL

** This is what I added
*STEP, NLGEOM
*STATIC
1, 1
*END STEP


*STEP, INC=1000000
*DYNAMIC,DIRECT,NLGEOM
 0.0001,0.01
*RESTART,WRITE,FREQUENCY=1
*BOUNDARY
 Ninlet,1,3,0
 Noutlet,1,3,0
*CLOAD
 Ninterface,1,0.0
 Ninterface,2,0.0
 Ninterface,3,0.0
*NODE FILE
 U
*EL FILE
 S, E
*END STEP
```

I get an error without this PR (as expected), but with a segfault during the static step. Any idea where it comes from ? Is it because I used a bad configuration or would it be a problem in the code ? (I assume having 2 coupled steps without a good usage of RESTART is doomed anyways)

MakisH:
Thanks a lot for the contribution! Are you also preparing a PR to the website/documentation?

MatthiasFreimuth:
I'm currently not preparing a PR to the website/documentation, but may do so if I find the time at some point.

boris-martin (approved):

KyleDavisSA (approved):

PR #102, opened 2022-07-24
Fixed crash when freeing unused pointers

boris-martin:
When one interface doesn't write data, an unitialized pointer gets freed in the end. This PR initalizes it to NULL when necessary.
Example to reproduce: https://github.com/precice/tutorials/pull/271.

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -700,6 +700,8 @@ void PreciceInterface_ConfigureCouplingData(PreciceInterface *interface, Simulat
   interface->numWriteData = config->numWriteData;
   if (config->numWriteData > 0)
     interface->writeData = malloc(config->numWriteData * sizeof(int));
+  else 
+    interface->writeData = NULL;
 
   for (i = 0; i < config->numWriteData; i++) {
     if (isEqual(config->writeDataNames[i], "Temperature")) {
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -700,7 +700,7 @@ void PreciceInterface_ConfigureCouplingData(PreciceInterface *interface, Simulat
   interface->numWriteData = config->numWriteData;
   if (config->numWriteData > 0)
     interface->writeData = malloc(config->numWriteData * sizeof(int));
-  else 
+  else
     interface->writeData = NULL;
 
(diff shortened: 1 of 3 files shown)


adapter/PreciceInterface.c
```diff
@@ -700,6 +700,8 @@ void PreciceInterface_ConfigureCouplingData(PreciceInterface *interface, Simulat
   interface->numWriteData = config->numWriteData;
   if (config->numWriteData > 0)
     interface->writeData = malloc(config->numWriteData * sizeof(int));
+  else
+    interface->writeData = NULL;
```
MakisH, line 704: I am not familiar with the codebase or sure about good practices in C, but I understand that will make sure that the free at line 769 is not triggered when no write-data is configured. I guess one could default this to `NULL` in the structure definition.

Shouldn't the same also apply to `readData`?

Applying this change does not fix the issue for me, neither in debug nor in release mode. Does this actually work for you? Does it work on the Demo VM or in some CI environment?

tools/format-code.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 # Format all adapter C and C++ files with clang-format
-find ./ \( -iname "*.h" -o -iname "*.hpp" -o -iname "*.c" -o -iname "*.cpp" \) -exec clang-format -i {} \;
+find ./ \( -iname "*.h" -o -iname "*.hpp" -o -iname "*.c" -o -iname "*.cpp" \) -exec clang-format-11 -i {} \;
```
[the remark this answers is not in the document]
boris-martin: Actually an another side-effect of using "git commit -a" at night, but we can keep it yes

adapter/PreciceInterface.c
boris-martin: To solve the error, I needed this PR and some changes to the tutorial case that I just pushed (I had forgotten yesterday): CalculiX time & preCICE end time were longer than OpenFOAM end time.
Does it work for you with this PR + updated tutorial?

(Might be interesting to generate better errors when time don't match, still).

I moved the change to initialization and also initialized `readData`.

MakisH (approved):
The simulation completes now and moving to the default value makes much more sense! :+1:

PR #103, opened 2022-08-10
Add error messages when reading invalid data in modal dynamic simulations

boris-martin:
Reading Dirichlet-like BCs (displacements, velocities, temperatures) doesn't work when using modal dynamic simulations, as they live in eigenspace and no conversion is done. Only loads (forces, pressures, heat fluxes) are expected to work.
This PR adds an error message when this is attempted, instead of silently not doing anything.


adapter/PreciceInterface.c
```diff
@@ -610,6 +610,19 @@ void PreciceInterface_EnsureValidNodesMeshID(PreciceInterface *interface, const
   }
 }
 
+void PreciceInterface_EnsureValidRead(SimulationData *sim, enum CouplingDataType type)
+{
+  // Forbidden read data in modal dynamic simulations
+  if (sim->isModalDynamic) {
+    if (type == TEMPERATURE || type == DISPLACEMENTS || type == DISPLACEMENTDELTAS || type == VELOCITIES || type == POSITIONS) {
+      printf("Error: in modal dynami simulations, only loads (forces, pressures, heat fluxes) can be read.\n"
```
MakisH, line 618:
```suggestion
      printf("Error: in modal dynamic simulations, only loads (forces, pressures, heat fluxes) can be read.\n"
```

adapter/PreciceInterface.h
```diff
@@ -141,15 +141,16 @@ typedef struct SimulationData {
   double  precice_dt;
   double  solver_dt;
 
+  // Configuration infos
```
MakisH, line 144:
```suggestion
  // Configuration information
```

dyna_precice.c
```diff
@@ -209,42 +209,43 @@ void dyna_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **lakonp
...
+      .xforc          = xforc,
+      .xboun          = xboun,
+      .ntmat_         = ntmat_,
+      .vold           = vold,
+      .veold          = veold,
+      .fn             = fn,
+      .cocon          = cocon,
+      .ncocon         = ncocon,
+      .mi             = mi,
+      .isModalDynamic = 1};
```
MakisH, line 248: Why so many changed lines? Is this just due to formatting? Looks ok, but just wondering.

MakisH (approved):
I understand the code changes and it looks mostly good. However, I did not try running it. To you want me to try to run and trigger this error?

dyna_precice.c
boris-martin: Yes, because the latest addition has a much longer name. Otherwise it's unchanged.

boris-martin:
> I understand the code changes and it looks mostly good. However, I did not try running it. To you want me to try to run and trigger this error?

I managed to trigger it myself, I don't think you need to bother :)

PR #104, opened 2022-08-10
Fixed DOF 0 bug & removed useless memcpy

boris-martin:
# Content

This PR seems to fix https://github.com/precice/calculix-adapter/issues/86
Following a suggestion from Matthias Freimuth, I removed a `memcpy` statement that seems to fixe the bug. Extensive testing is needed to ensure nothing else is broken. For the record, I am absolutely clueless as to 1) why removing this line fixes the bug 2) what was the purpose of it at first. So, mostly empirical stuff.

In my opinion, three things should be checked:
- Do all the tutorials work correctly with this change?
- Does this change fix the #86 bug wherever it occurs?
- Does this change impact memory consumption or something related? (I don't think so, but worth monitoring I guess)

@KyleDavisSA any idea what's happening here ?

# Checklist

On my side:

## Fixing the bug?

- [x] Perpendicular flap modified as suggested by Andres Pedemontes (see #86 ) is fixed.
- [x] The test case I introduced at https://github.com/precice/calculix-adapter/pull/92#issuecomment-1061668873 is fixed.

## Non-regression

- [x] Does the perpendicular flap work as before ?
- [x]  Does the heat exchanger work as before ?
- [x]  Does the elastic-tube-3d work as before ?
- [x]  Does the flow over heated plate work as before ?
- [ ]  Does the partitioned beam work as before ?

## Memory checks

Nothing done so far.

### DIFF
--- a/nonlingeo_precice.c
+++ b/nonlingeo_precice.c
@@ -3694,7 +3694,6 @@ void nonlingeo_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **l
 
     /* Adapter: Copy necessary data for coupling */
     simulationData.fn = fn;
-    memcpy(&vold[0], &v[0], sizeof(double) * mt * *nk);
 
     Precice_WriteCouplingData(&simulationData);
     /* Adapter: Advance the coupling */


MakisH (commented):
I also don't understand what this line is for. I suspect it was a leftover from early development, more of a "safety measure" before replacing the `vold` values in the upcoming `read`. It does look useless to me with my very limited understanding. I assume that something is wrong with the sizes, and I would not be surprised if this is a Fortran/C interoperability issue (in Fortran, indices start from 1, in C from 0). `memcpy` is very unsafe, but I don't think we have better options here.

Running the perpendicular flap case does seem to trigger this part of the code (I added a `printf` to quickly check), and the result files are identical.

I would say let's merge and wait for complaints. And definitely inform the people that reported this before (also on Discourse).

I found the same operation commented-out on line 3934:

That one was there since the beginning, and commented-out by @KyleDavisSA when he upgraded to v2.16: [git blame old state](https://github.com/precice/calculix-adapter/blame/505e90aa736c410da41bfd3578daa0bef7989f64/nonlingeo_precice.c) (line 960), [git blame new state](https://github.com/precice/calculix-adapter/blame/d9c8bc75e242e6480d8946ddc5acf2eeb0ab7c9a/nonlingeo_precice.c) (line 1190). I understand that then he moved it to https://github.com/precice/calculix-adapter/blame/a02ff512d12f5ceb57a39aaca3c839258547600b/nonlingeo_precice.c#L3457 (but I may misunderstand this step of the adapter evolution).

MakisH:
@MatthiasFreimuth could you please help us understand this bugfix?

@boris-martin the "Does the partitioned beam work as before ?" box is still unticked. Does this mean you still want to test it, or that you saw issues?

If you say that this already fixed some known cases and it does not break others, let's merge it.

boris-martin:
I didn't test it yet. But given the good results on other cases I don't expect any issue.

Issue #9, opened 2018-09-29
Checkpointing does not work for subcycling


derekrisseeuw:
For implicit coupling the writeCheckpoint is called by the function:
`Precice_WriteIterationCheckPoint`

This function takes the values of the last timestep as checkpoint values. However, for implicit coupled simulations where the calculix is subcycling, the last calculix timestep is not the same as the last coupling timestep. Therefore, the wrong checkpoint is written.

uekerman:
Thanks for pointing us to this. I guess that this has simply never been implemented.

derekrisseeuw:
Thank you for your reaction Benjamin. 

This problem can be easily overcome in the calculix input file by specifying: 
*DYNAMIC,DIRECT
<timestep>

Where the 'direct' keyword forces calculix to use the timestep specified (this should be the precice timestep).

nkr0:
Doesn't putting write/read checkpoints inside `if *required` prevent this? My understanding is that the solver has to be at one of the ends of a coupling step for these to return true.

If that is the case, CalculiX sub-cycles by storing states in `vini` and `vold` without touching `simulationData->coupling_init_v`, which is only written/read at the beginning/end of coupling steps. precice resets `vold` to the beginning of the coupling time step, when `Precice_IsReadCheckpointRequired` returns true, i.e., only when the coupling step is not converged at the end of a coupling step (not at the end of CalculiX sub-cycles). Similarly, `Precice_IsWriteCheckpointRequired` returns true only at the beginning of a coupling time step to store a checkpoint in `simulationData->coupling_init_v`.

https://github.com/precice/calculix-adapter/blob/b01641e40c11e067f66cc41fbcb0253d08290ff0/nonlingeo_precice.c#L1115-L1121
https://github.com/precice/calculix-adapter/blob/b01641e40c11e067f66cc41fbcb0253d08290ff0/nonlingeo_precice.c#L2664-L2678

`Read/WriteCouplingData`, `AdjustSolverTimestep`, and `Advance` executes every sub-cycle, to
https://github.com/precice/calculix-adapter/blob/b01641e40c11e067f66cc41fbcb0253d08290ff0/nonlingeo_precice.c#L1097-L1099
https://github.com/precice/calculix-adapter/blob/b01641e40c11e067f66cc41fbcb0253d08290ff0/nonlingeo_precice.c#L2666
read/write `simulationData->preciceInterfaces`,
[...]

PR #105, opened 2022-08-14
Fixed subcycling in implicit modal dynamic simulations

boris-martin:
Partial fix to https://github.com/precice/calculix-adapter/issues/9 (which was first noticed for non-modal dynamic simulations)
Tried on the perpendicular flap modified to have smaller steps than the time window size. (I tried 0.5 * window size and 0.2. Non-exact dividers like 0.4 might be worth trying too).

I'm still confused by the output frequency: in the current state, it outputs only once per time window instead of once per step, and there is no extra writes from removed iterations. I'm very confused by the role of `iinc` and `jprint`, and I can't tell if the write happens after the first step of an iteration or at the end of it. Investigation still ongoing.

Result-wise, it's quite convincing and seems to match correctly the result with bigger steps.

I also noticed that with subcycling I get one less VTK step outputted than with full steps.


MakisH:
What would you like me to review, exactly? Because I still get a mesh failure in OpenFOAM (but I assume this is what you meant with "partial fix").

Using a timestep size of 3e-3 for the perpendicular flap, I get 163 steps. Applying a "warp by vector" (with displacement) filter on the CalculiX results (converted from FRD), the beam keeps getting longer, until it breaks down to an arbitrary shape.

> I'm still confused by the output frequency: in the current state, it outputs only once per time window instead of once per step, and there is no extra writes from removed iterations. I'm very confused by the role of iinc and jprint, and I can't tell if the write happens after the first step of an iteration or at the end of it. Investigation still ongoing.

I remember trying a similar investigation for the OpenFOAM adapter, where I used the `watch` tool to monitor changes in the results files. Maybe that helps.

> Result-wise, it's quite convincing and seems to match correctly the result with bigger steps.

Does this mean that this fixes #9? Because for me, it does not look like it.

> I also noticed that with subcycling I get one less VTK step outputted than with full steps.

This worries me more. Maybe now CalculiX writes only in the beginning and not at the end? Maybe you can compare result files and see if they correspond at the same time.

I have not looked at the code yet, but I don't think I have enough knowledge of it to review it.

boris-martin:
> What would you like me to review, exactly? Because I still get a mesh failure in OpenFOAM (but I assume this is what you meant with "partial fix").

Actually, not that much now, mostly aesthetics (does it make sense to store the watchpoint in that data structure? Would a separate structure be cleaner ? Etc)

> Using a timestep size of 3e-3 for the perpendicular flap, I get 163 steps. Applying a "warp by vector" (with displacement) filter on the CalculiX results (converted from FRD), the beam keeps getting longer, until it breaks down to an arbitrary shape.

> Does this mean that this fixes #9? Because for me, it does not look like it.

Did you run a modal simulation? (`run.sh -modal`) Because the regular simulations are not fixed yet (hence the partial fix). Both kind must be fixed but this one is slightly easier (I've worked more on it, and the file is only 2000 lines long instead of 4000 :see_no_evil: ), which is why I'm starting on this side. But fixing non-modal dynamic (i.e. regular) simulations will take another PR.

> > I'm still confused by the output frequency: in the current state, it outputs only once per time window instead of once per step, and there is no extra writes from removed iterations. I'm very confused by the role of iinc and jprint, and I can't tell if the write happens after the first step of an iteration or at the end of it. Investigation still ongoing.
> 
> I remember trying a similar investigation for the OpenFOAM adapter, where I used the `watch` tool to monitor changes in the results files. Maybe that helps.

I don't quite understand the approach. What would I gain compared to logging the times where the `frd()` function (the only one writing output) is called?


> > I also noticed that with subcycling I get one less VTK step outputted than with full steps.
> 
> This worries me more. Maybe now CalculiX writes only in the beginning and not at the end? Maybe you can compare result files and see if they correspond at the same time.

First steps are matching but an extra step is added.


> I have not looked at the code yet, but I don't think I have enough knowledge of it to review it.

An opinion of @KyleDavisSA  would indeed be welcome.

MakisH:
You are right, I forgot to enable the modal simulation mode. In that case, it looks like it solves the issue! :smile: 

> I don't quite understand the approach. What would I gain compared to logging the times where the frd() function (the only one writing output) is called?

Probably none. Indeed, different approaches are easier in this case.

dyna_precice.c
```diff
@@ -2039,9 +2047,9 @@ void dyna_precice(double **cop, ITG *nk, ITG **konp, ITG **ipkonp, char **lakonp
     /* Adapter: If the coupling does not converge, read the checkpoint */
     if (Precice_IsReadCheckpointRequired()) {
       if (*nmethod == 4) {
-        Precice_ReadIterationCheckpoint(&simulationData, vold);
-        memcpy(&bj[0], &cd[0], sizeof(double) * nev);
-        memcpy(&bjp[0], &cv[0], sizeof(double) * nev);
+        Precice_ReadIterationCheckpointModal(&simulationData, bj, bjp, nev);
+        //memcpy(&bj[0], &cd[0], sizeof(double) * nev);
+        //memcpy(&bjp[0], &cv[0], sizeof(double) * nev);
```
MakisH, line 2052: Please comment on why you comment-out such things, or remove them completely. We discussed previously such a cryptic commented-out line. :see_no_evil:

adapter/PreciceInterface.h
```diff
@@ -216,6 +222,24 @@ void Precice_ReadIterationCheckpoint(SimulationData *sim, double *v);
  */
 void Precice_WriteIterationCheckpoint(SimulationData *sim, double *v);
 
+/**
+ * @brief Reads iteration checkpoint (in dyna_precice)
```
MakisH, line 226: We need a comment for the motivation for the duplication here (and in the writing method). What is special in modal simulations?

adapter/PreciceInterface.c
```diff
@@ -160,6 +160,40 @@ void Precice_WriteIterationCheckpoint(SimulationData *sim, double *v)
...
+
+  // Reload time
+  *(sim->theta) = sim->coupling_init_theta;
+
+  // Reload step size
+  *(sim->dtheta) = sim->coupling_init_dtheta;
+
+  // Reload DOFs in eigenmodes space & counters
+  memcpy(dofs, sim->eigenDOFs, sizeof(double) * nev);
+  memcpy(derivatives, sim->eigenDOFsDerivatives, sizeof(double) * nev);
```
MakisH, line 177: Wouldn't it make more sense to add these in the main read/write checkpoint methods and only enable them if this is a modal simulation?

dyna_precice.c
boris-martin: Cleaning oversight again 😅

adapter/PreciceInterface.h
boris-martin: Saved data is different. The saved data in modal simulations doesnt exist in regular ones.

But an if(isModal) inside would work too. Except that the number of arguments changes too, so I'm not sure

adapter/PreciceInterface.c
boris-martin: That would probably mess up the arguments.unless i move the calculix data into the SimData, but that would be a bit messy and a break of separation of concerns in my opinion

adapter/PreciceInterface.h
MakisH: Great motivations, let's add it to a comment!

adapter/PreciceInterface.c
MakisH: If you feel that this would be more complicated, we can also leave it like this and look at it again in the future.

boris-martin:
I've got a running draft of a change that solves the output issue. Since we can't override previously written output, I started storing intermediate results, then at the end of a time window that converged (`precicec_IsTimeWindowComplete()`), we write all the stored results at once. If the time window is started again (a checkpoint is read), then we clear the buffer.

Since CalculiX is pretty much a blackbox in terms for understanding the output code, I designed a flexible buffer to handle changes if needed, as I didn't know exactly what had to be stored. The code is still a bit messy, and I'd like some conceptual approval on it before investing time on polishing the details.

This buffer code should be reusable for the non-modal dynamic code later :)

MakisH (commented):
With the updated state:
- I get 500 output files instead of 163
- The modal case works, the non-modal remains the same

Regarding the big picture, I think the need and mechanics of the OutputBuffer need to be better documented. Since CalculiX does not expect to go back in time, I can understand that there is such a need for an OutputBuffer specifically for the adapter.

Of course, if there is a simpler way, it would be nice to save the additional complexity.

adapter/OutputBuffer.hpp
```diff
@@ -0,0 +1,47 @@
+/**********************************************************************************************
+ *                                                                                            *
+ *       CalculiX adapter for heat transfer coupling and mechanical FSI using preCICE         *
+ *       Heat transfer adapter developed by Lucía Cheung with the support of SimScale GmbH    *
+ *                                                                                            *
+ *       Adapter extended to fluid-structure interaction by Alexander Rusch                   *
+ *                                                                                            *
+ *********************************************************************************************/
+
+#ifndef CCX_OUTPUT_BUFFER_HPP
```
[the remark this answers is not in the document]
boris-martin: The main code is in C, so we can't import the .hpp file. It only exists to provide a C/C++ "binding" where the cpp file implements OOP code declared in hpp *and* the C API

adapter/OutputBuffer.h
```diff
@@ -0,0 +1,38 @@
...
+ *       Adapter extended to fluid-structure interaction by Alexander Rusch                   *
+ *                                                                                            *
+ *********************************************************************************************/
+
+#ifndef CCX_OUTPUTBUFFER_H
+#define CCX_OUTPUTBUFFER_H
+
+// To avoid including CalculiX.h
+#ifdef LONGLONG
+#define ITG long long
```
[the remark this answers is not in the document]
boris-martin: I guess it's "integer" since it's either int or long, so it's a macro to accept 64-bits indexes.
Not sure it's documented anywhere, it's mostly a guess.

adapter/PreciceInterface.h
```diff
@@ -149,6 +149,7 @@ typedef struct SimulationData {
   double *eigenDOFsDerivatives; // Called "bjp" (p for prime) or "cv" in CCX code
   int     stored_iinc;
   int     stored_jprint;
+  int     kode_value;
```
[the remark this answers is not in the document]
boris-martin: It's the stored value of the CalculiX `kode` variable. Which is very ill-named as it is indeed more of a convergence status / iteration counter thing, but I tried to remain faithful to the original naming (which I don't understand myself, TBH).

boris-martin:
> Regarding the big picture, I think the need and mechanics of the OutputBuffer need to be better documented.

I see how its internals should be documented, but where would you document the need? Inside the dyna_precice.c code ? That's actually why I left the commented `frd` call, to remember its "old location", but it's probably a good idea to put more text there.

KyleDavisSA:
I tested this PR for the perpendicular-flap case with various sub-cycling. If I remember correctly, it was failing when one time step was not a multiple of the other. I had no problems this time around. Am I remembering this correctly.

Also, where can I find an example for the modal dynamic case?

boris-martin:
This should produce a correct output with all step sizes, but with visualization artifact (at least in ParaView with ccx2paraview as converter), as every outputted thing is seen as a unit step.

Modal case is the perpendicular flap with `./run.sh -modal`

KyleDavisSA (approved):

PR #106, opened 2022-09-25
Fixed subcycling in implicit coupling (for regular simulations)

### LINKED ISSUE #9: Checkpointing does not work for subcycling

boris-martin:
(Depends on #105 )

Fixes https://github.com/precice/calculix-adapter/issues/9. Essentially, checkpointing stores more data than previously (not only positions, but velocities and a few more things), because those additional fields must be stored when going back from more than an increment (time step). So previous implementation worked only when there was one step per window.

Still WIP:

- [x] Refactoring & cleaning
- [ ] Outputting once per time step instead of once per window (using the buffer from #105)
- [ ] Allow user-defined output frequency (once every N steps)
- [ ] Testing on more cases, including thermal cases (only perpendicular flap so far)
- [x] Check for redundancy (I may have copied more data than necessary?)

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -443,6 +443,14 @@ void Precice_FreeData(SimulationData *sim)
     free(sim->coupling_init_v);
   }
 
+  if (sim->coupling_init_ve != NULL) {
+    free(sim->coupling_init_ve);
+  }
+
+  if (sim->coupling_init_acc != NULL) {
+    free(sim->coupling_init_acc);
+  }
+
   for (i = 0; i < sim->numPreciceInterfaces; i++) {
     PreciceInterface_FreeData(sim->preciceInterfaces[i]);
     free(sim->preciceInterfaces[i]);
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -51,6 +51,9 @@ void Precice_Setup(char *configFilename, char *participantName, SimulationData *
 
   // Initialize variables needed for the coupling
   NNEW(sim->coupling_init_v, double, sim->mt * sim->nk);
+  NNEW(sim->coupling_init_ve, double, sim->mt * sim->nk);
+  NNEW(sim->coupling_init_acc, double, sim->mt * sim->nk);
+  NNEW(sim->coupling_init_xboun, double, sim->nboun);
 
   // Initialize preCICE
   sim->precice_dt = precicec_initialize();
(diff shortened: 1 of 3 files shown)


boris-martin:
Turns out adapting the buffer code is trickier than anticipated. I'm undrafting and I suggest staying on the current partial fix: correct results, but with one output every N time window instead of every N steps, with N defaulting to 1.
This will limit the introduction of new bugs for now.

KyleDavisSA (approved):
Changes look good. All new variables are freed in the end.  However the perpendicular flap does not function correctly. 

- develop branch with equal time steps functions correctly, but breaks for subcycling.
- with this PR, subcycling and equal time steps do not crash, but the flap hardly moves.

MakisH (changes requested):
I tried running the perpendicular flap, but it looks like we need to consider updates based for v2.20:

- With v2.19 and this PR, the tutorial was running normally
- With v2.20, it is running normally
- With v2.20 and this PR, the flap does not move

Please also check if it actually fixes the original issue ([example on how to reproduce it](https://github.com/precice/calculix-adapter/issues/9#issuecomment-1199067524)).

KyleDavisSA:
My suggestion would be to continue with the v2.20 release and continue to work on subcycling. Rather have the adapter working correctly when all time steps are equal.

MakisH:
> > I tried running the perpendicular flap, but it looks like we need to consider updates based for v2.20:
> > ```
> > * With v2.19 and this PR, the tutorial was running normally
> > 
> > * With v2.20, it is running normally
> > 
> > * With v2.20 and this PR, the flap does not move
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please also check if it actually fixes the original issue ([example on how to reproduce it](https://github.com/precice/calculix-adapter/issues/9#issuecomment-1199067524)).
> 
> How did you build "This + 2.20 ?" Since this PR involves quite some modifications of the main file I planned to do it by hand

I checked out the `v2.20` branch and I merged this branch, which did not lead to any conflicts. I guess we do need to update things, but let's make that a bugfix release, then. There is no regression.

MakisH:
This now needs to be rebased on the latest state of the adapter, which has two version changes:

- CalculiX v2.19 -> v2.20
- preCICE v2 -> v3

Since the changes in this PR are relatively few, I suggest applying the ideas from this PR directly in the latest state of the code.

Since @boris-martin is not working on this anymore, I am closing this. If anyone wants to pick this up, start a new PR from develop, apply the same ideas, and add @boris-martin as co-author.

PR #108, opened 2022-10-24
Make read/write data name parsing consistent

### LINKED ISSUE #107: Inconsistent lookup for read/write data in CHT and FSI
For temperature and heat flux, we are checking whether the field name is exactly matched in the config:

https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/PreciceInterface.c#L624

For other fields, we are checking whether the field name just starts with this prefix (similar behavior as in the OpenFOAM adapter), sometimes with a trailing `-`, sometimes without:

https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/PreciceInterface.c#L37

https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/PreciceInterface.c#L654

Additionally, data names are hard-coded for preCICE:

https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/PreciceInterface.c#L628

Finally, this error message is rather confusing:

https://github.com/precice/calculix-adapter/blob/4496e8f160ec121481653baf44ec032813409ac6/adapter/PreciceInterface.c#L662

"Does not exist" made me think that I still needed to specify it somewhere. However, it just means that the adapter does not know how to handle it.

MakisH:
Fixes #107.

- Make names parsing consistent for all read/write data
- Improve error message for unknown types

Dear reviewer: have I forgotten anything? Have I covered all cases? (not sure who would like to have a look on that)

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -661,23 +661,23 @@ void PreciceInterface_ConfigureCouplingData(PreciceInterface *interface, Simulat
   if (config->numReadData > 0)
     interface->readData = malloc(config->numReadData * sizeof(int));
   for (i = 0; i < config->numReadData; i++) {
-    if (isEqual(config->readDataNames[i], "Temperature")) {
+    if (startsWith(config->readDataNames[i], "Temperature")) {
       PreciceInterface_EnsureValidNodesMeshID(interface, "Temperature");
       PreciceInterface_EnsureValidRead(sim, TEMPERATURE);
       interface->readData[i]       = TEMPERATURE;
       interface->xbounIndices      = malloc(interface->numNodes * sizeof(int));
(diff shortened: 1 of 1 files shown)


boris-martin:
I think the "startsWith" is necessary to handle multiple access to the same thing (Temperature-1, Temperature-2), but not sure anymore. @IshaanDesai has worked on that part I think ?

boris-martin:
My bad I misread,you changed the `isEqual` to `startsWith`, not the other way around. So it should be good I guess. Might be worth checking all the tutorials still pass

KyleDavisSA (approved):

boris-martin (approved):

PR #109, opened 2022-11-15
New branch for v2.20 pre-release

KyleDavisSA:
This PR starts the procedure for the v2.20 release.


boris-martin:
I assume we want to first merge the subcycling related PRs before actually updating ?

KyleDavisSA:
PR #105 right? PR #106 is still WIP.

boris-martin:
I'd actually merge #106 almost as-is (so it fixes the bug but we have an output per time window instead of per step). It depends on 105 (If I remember right, it's been some time), but no major update is needed, AFAIK

MakisH:
Please also remember to update the documentation: https://precice.org/adapter-calculix-overview.html

MakisH:
Note that a few CI checks seem to be failing at the moment.

KyleDavisSA:
Perpendicular flap works for equal time step sizes, but not sub-cycling (like v2.19). The sub-cycling can be addressed within v2.20 as that has not been solved yet.

MakisH (approved):
The perpendicular-flap OpenFOAM-CalculiX also runs for me. The code changes look reasonable.

PR #115, opened 2023-11-01
Port adapter to CCX v2.21

numericalfreedom:
Update to CalculiX version 2.21 (July 2023)


uekerman:
Thanks for the important contribution @numericalfreedom :rocket:

numericalfreedom:
Dear Benjamin,

thank You for Your kind comments.

The preCICE version of ccx_2.21 compiles and works fine with all native CalculiX tests (standalone ccx_preCICE works fine).

Between the versions ccx_2.20 and ccx_2.21 there were only some cosmetic changes as far as it is concering the preCICE connection. I assume, that the version ccx_2.21 with preCICE works also fine but this remains to be thoroughly tested. I am too much a newbee to have several codes available to couple CalculiX with and also to have the detailed knowledge to do so. Maybe, I can use Your expertise on this.

Thank You very much for Your kind attention and best wishes,

Nandor
@numericalfreedom on Github

Makefile
```diff
@@ -2,24 +2,33 @@
 # https://precice.org/adapter-calculix-get-calculix.html
 # Set the following variables before building:
 # Path to original CalculiX source (e.g. $(HOME)/ccx_2.xx/src )
-CCX_VERSION			= 2.20
-CCX             = $(HOME)/CalculiX/ccx_$(CCX_VERSION)/src
+CCX_VERSION	= 2.21
+CCX_HOME        = /opt/src
+CCX             = $(CCX_HOME)/CalculiX/ccx_$(CCX_VERSION)/src
```
MakisH, line 7: The `CCX_VERSION` should indeed be updated, but the `CCX` should remain as-is.
I don't think we need a separate `CCX_HOME` variable, it makes copy-pasting paths more complicated. If it helps further parts or CalculiX-internal workflows, let's keep it.

Makefile
```diff
@@ -53,6 +62,7 @@ LIBS = \
 #FFLAGS = -g -Wall -O0 -fopenmp $(INCLUDES)
 
 CFLAGS = -Wall -O3 -fopenmp $(INCLUDES) -DARCH="Linux" -DSPOOLES -DARPACK -DMATRIXSTORAGE -DUSE_MT
+CFLAGS =  -w  -O3 -fopenmp $(INCLUDES) -DARCH="Linux" -DSPOOLES -DARPACK -DMATRIXSTORAGE -DUSE_MT
```
MakisH, line 65: Why changing from `-Wall` to `-w`?
If it was intended to add the `-Wextra`, then let's just add it explicitly.

Or is this a change that comes from the CalculiX v2.21 Makefile?

Same below.

ccx_2.21.c
```diff
@@ -0,0 +1,1923 @@
+/*     CalculiX - A 3-dimensional finite element program                 */
```
MakisH, line 1: Comparing this file with `ccx_2.20.c`, it looks like this file does not contain the preCICE-specific changes. Did you maybe accidentally replace the complete file with the unmodified Calculix?

MakisH (commented):
Thank you for the update!

The Makefile needs some attention, but I managed to build the adapter using the same Makefile as in ccx2.20 (with only the version changed).

Besides that, I have the impression that the files `ccx_2.21.c` and `CalculiX.h` are the unmodified CalculiX files, not the adapted ones. I noticed that by trying to run a tutorial, in which case the CalculiX participant just started iterating without waiting for the other participant.

Let me know if you need any help applying my suggestions! :-)

MakisH:
Closing this, since there has been no reaction to the review points and there are already newer CalculiX versions, which we should directly port to.

PR #119, opened 2024-01-30
Use includes in more build steps

MakisH:
Without these, I cannot build with preCICE in the build directory, without modifying my `INCLUDE_PATH` etc. I never could, but now with v2.5.1, I was sure that it should somehow work only with pkgconfig.


IshaanDesai (approved):
Looks good :+1:

PR #123, opened 2024-04-01
Remove last reference to libprecice2 in the Debian packaging recipe

IshaanDesai:


MakisH (approved):
I somehow did not see this earlier. Makes sense! :+1:

PR #124, opened 2024-04-05
Update dependency in debian package recipe

IshaanDesai:
This release is necessary because an old reference to `libprecice2` was leftover in the previous release [v2.20.1](https://github.com/precice/calculix-adapter/releases/tag/v2.20.1). Fix in https://github.com/precice/calculix-adapter/pull/123


.github/workflows/ubuntu_build.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     container: precice/precice:develop
     continue-on-error: true
     env: # Versioning is "calculix-preciceX_2.YY-Z[...]" with X the major preCICE version, YY the minor CCX version and Z the package version
-      PACKAGE_NAME: "calculix-precice3_2.20.1-1_amd64"
```
MakisH, line 23: Since we are only changing the Debian package details, the most appropriate here would be `2.20.1-2`. But it probably does not matter anywhere.

MakisH (approved):

.github/workflows/ubuntu_build.yml
IshaanDesai: I changed this because we are bumping the version and doing a new release. `2.20.1` is the actual adapter version, right?

MakisH: - v2.20: CalculiX version
- v2.20.1: The second adapter we release that is made for v2.20 (previous was v2.20.0)
- v2.20.1-1: The first Debian package release (revision) of v2.20.1
- v2.20.1-2: The second Debian package release (revision) of v2.20.1.

IshaanDesai: By this logic we should not release. Because we are not changing the adapter, but just a dependency of the Debian package.

MakisH: I would actually prefer that. Let's rename this PR, still merge it, and update the Debian packages in the release with a note, in case the previous ones were broken.

IshaanDesai:
I renamed the branch locally and pushed it and then deleted the old branch, which led GitHub to automatically close the pull request. I thought it would move the base branch to the new one, but apparently this does not work. Now I opened a new PR: https://github.com/precice/calculix-adapter/pull/126

PR #126, opened 2024-04-08
Update libprecice dependency in the debian package recipe

IshaanDesai:
Updating an old reference to `libprecice2` was leftover in the previous release [v2.20.1](https://github.com/precice/calculix-adapter/releases/tag/v2.20.1). Fix in https://github.com/precice/calculix-adapter/pull/123


IshaanDesai:
@MakisH do you think it is appropriate that we merge this directly to master? Currently the target branch is develop.

MakisH (approved):

PR #118, opened 2024-01-17
Add functionality for volumetric coupling of material models

kalupaika:
This PR adds functionality to do volumetric coupling for mutliscale structural simulations. CalculiX is used on the macro or meso scale, and is coupling via preCICE and the Micro Manager to micro-mechanics solvers.

This PR is opened mainly to investigate the necessary changes to the adapter and is not in a state of being merged directly.

PR #130, opened 2024-11-06
Update actions/upload-artifact version

IshaanDesai:
This PR updates the used versions of https://github.com/pre-commit/mirrors-clang-format and actions/upload-artifact to more newer versions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 # Official repo for the clang-format hook
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: 'v19.1.2'
+  rev: 'v8.0.1'
   hooks:
   - id: clang-format
     files: "^adapter"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 # Official repo for the clang-format hook
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: 'v8.0.1'
+  rev: 'v19.1.2'
   hooks:
   - id: clang-format
     files: "^adapter"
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 repos:
 # Official repo for the clang-format hook
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: 'v8.0.1'
+  rev: 'v19.1.2'
```
MakisH: This upgrade would mean reformatting the codebase, which is not triggered here because no C++ files are changed in this PR.

What is the motivation for updating clang-format?

IshaanDesai: Due to the older version of clang-format, I hit the following error in Ubuntu 24.04:

```bash
clang-format.............................................................Failed
- hook id: clang-format
- exit code: 1

Traceback (most recent call last):
  File "/home/desaiin/.cache/pre-commit/repops94iqy7/py_env-python3.12/bin/clang-format", line 5, in <module>
    from clang_format import clang_format
  File "/home/desaiin/.cache/pre-commit/repops94iqy7/py_env-python3.12/lib/python3.12/site-packages/clang_format/__init__.py", line 5, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```
This is resolved by running `/home/desaiin/.cache/pre-commit/repops94iqy7/py_env-python3.12/bin/pip install setuptools` because setuptools is missing in the backend venv that pre-commit uses. Upgrading would resolve this issue, which we will hit more often than not.

I understand that this will heavily impact the code base, which would result in a lot of friction for some pull requests. But we should do it at some point.

MakisH: I do not disagree with this, I just think we should do it in another PR to get this merged first.

IshaanDesai: Okay, we can do it in another PR. It would also make sense to merge https://github.com/precice/calculix-adapter/pull/118 before updating clang-format.

MakisH:
```suggestion
  rev: 'v8.0.1'
```

MakisH (approved):

PR #131, opened 2024-11-07
Run pre-commit via GitHub Actions

IshaanDesai:
This PR does the following things:
1. Removes the Action to check formatting via clang-format, because it is now checked by directly running pre-commit. This increases maintainability as the clang-format version is now defined at only one place, the pre-commit config file.
2. Runs pre-commit via an Action, which checks formatting and also other things that are included in the pre-commit checks
3. The Ubuntu build Action now runs only once in a pull request to the main or develop branches. Earlier there used to always be two runs per push via a pull request.


MakisH (approved):
Makes sense to me, thanks for tidying up!

PR #135, opened 2025-03-20
Add functionality for volumetric multiscale coupling to micro-mechanics simulations

IshaanDesai:
This PR is derived from #118 and adds functionality for volumetric coupling to micro-mechanics simulations. The functionality is primarily contributed by @kalupaika. This PR is branched from the feature branch of #118 because #118 contains code specific to coupling a macro-scale CalculiX simulation with many micro-scale [NASMAT](https://ntrs.nasa.gov/citations/20230015286) simulations. This PR contains the functionality:

- Get strain tensor from CalculiX and write it to preCICE.
- Read stress tensor components from preCICE and apply them in CalculiX.
- Read stiffness matrix components from preCICE and apply them in CalculiX.

The implemented functionality is tested only for steady-state simulations. To incorporate the features for transient simulations, an appropriate test case would need to be devised.


Makefile
```diff
@@ -62,7 +62,7 @@ else
 	CC = mpicc
 endif
 
-FFLAGS = -Wall -O3 -fopenmp $(INCLUDES) ${ADDITIONAL_FFLAGS}
+FFLAGS = -Wall -O3 -fopenmp -fallow-argument-mismatch $(INCLUDES) ${ADDITIONAL_FFLAGS}
```
IshaanDesai: This change needs to be removed before merging.

Makefile
```diff
@@ -73,8 +73,8 @@ include $(CCX)/Makefile.inc
 SCCXMAIN = ccx_$(CCX_VERSION).c
 
 # Append additional sources
-SCCXC += nonlingeo_precice.c dyna_precice.c CCXHelpers.c PreciceInterface.c
-SCCXF += getflux.f getkdeltatemp.f
+SCCXC += nonlingeo_precice.c dyna_precice.c CCXHelpers.c PreciceInterface.c linstatic_precice.c
+SCCXF += getflux.f getkdeltatemp.f getc3d8elementgausspointcoords.f getc3d4elementgausspointcoords.f
```
MakisH: Do I understand correctly that these files are completely written for the adapter, and not included in the original CalculiX source code? In that case, I wonder if they should better be under `adapter/` or another new folder. What is the motivation for putting them outside?

The operations they define are specific to the adapter, right?

adapter/CCXHelpers.c
```diff
@@ -68,6 +68,17 @@ void getSurfaceElementsAndFaces(ITG setID, ITG *ialset, ITG *istartset, ITG *ien
...
 
+void getElementsIDs(ITG setID, ITG *ialset, ITG *istartset, ITG *iendset, ITG *elements)
+{
+
+  ITG i, k = 0;
+
+  for (i = istartset[setID] - 1; i < iendset[setID]; i++) {
+    elements[k] = ialset[i];
+    k++;
+  }
```
MakisH, line 78: Do we need some kind of bound check for `k`? What is the size of `elements` supposed to be?

adapter/CCXHelpers.c
```diff
@@ -409,6 +420,18 @@ int getXloadIndexOffset(enum xloadVariable xloadVar)
   }
 }
 
+void getElementStrain(int strainIdx, int numIPTotal, double *eei, double *strainData)
```
MakisH: This function uses `int` for ids and sizes, but other functions use `ITG`. What is `ITG`, and is it a better-fitting type?

Same for the other two new functions.

adapter/CCXHelpers.c
```diff
@@ -409,6 +420,18 @@ int getXloadIndexOffset(enum xloadVariable xloadVar)
...
+{
+  int i, idx;
+  // Loop through all element and respective gauss points
+  for (i = 0; i < numIPTotal; i++) {
+    idx                   = i * 6 + strainIdx; //TODO: Add explanation for 6
```
MakisH, line 418: yes, please. Two similar points below.

adapter/CCXHelpers.h
```diff
@@ -241,6 +262,15 @@ void getXbounIndices(ITG *nodes, ITG numNodes, int nboun, int *ikboun, int *ilbo
  */
 void getXforcIndices(ITG *nodes, ITG numNodes, int nforc, int *ikforc, int *ilforc, int *xforcIndices);
 
+/**
+ * @brief Gets the strain values at each Gauss point of each element
+ * @param strainIdx: CalculiX variable for the index of the strain values
+ * @param numIPTotal: CalculiX variable for the number of elements
+ * @param eei: CalculiX array for the element integration information
```
MakisH: What does `eei` stand for? Is there a more intuitive name?
I guess this comes from some CalculiX-internal variable name, but I am still wondering.

adapter/PreciceInterface.h
```diff
@@ -277,6 +302,13 @@ void sendFaceCentersVertices(PreciceInterface *interface);
  */
 void PreciceInterface_ConfigureNodesMesh(PreciceInterface *interface, SimulationData *sim);
 
+/**
+ * @brief Configures the elements mesh and calls setMeshVertices on preCICE
+ * @param interface
+ * @param sim
```
MakisH: Descriptions missing

ccx_2.20.c
```diff
@@ -1308,7 +1308,7 @@ int main(int argc, char *argv[])
     /* nmethod=15: Crack propagation */
     /* nmethod=16: Feasible direction based on sensitivity information */
     if (preciceUsed) {
-      int isStaticOrDynamic = ((nmethod == 1) || (nmethod == 4)) && (iperturb[0] > 1);
+      int isStaticOrDynamic = ((nmethod == 1) || (nmethod == 4)) ;//&& (iperturb[0] > 1); Necessary to trigger isStaticorDynamic
```
MakisH: I don't understand what is going on here. What was the previous check, and why is it now removed? Do we want to keep the commented-out version for some reason?

ccx_2.20.c
```diff
@@ -1432,6 +1432,55 @@ int main(int argc, char *argv[])
...
+
+      } else if (isStaticOrDynamic) {
+
+        mpcinfo[0] = memmpc_;
+        mpcinfo[1] = mpcfree;
+        mpcinfo[2] = icascade;
+        mpcinfo[3] = maxlenmpc;
+
+        if (icascade != 0) {
+          printf(" *ERROR in CalculiX: the matrix structure may");
```
MakisH: The error is from the adapter, not from CalculiX itself, right?
In other places, we just write `ERROR`.

getc3d4elementgausspointcoords.f
```diff
@@ -0,0 +1,86 @@
...
+!     This program is distributed in the hope that it will be useful,
+!     but WITHOUT ANY WARRANTY; without even the implied warranty of 
+!     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+!     GNU General Public License for more details.
+!
+!     You should have received a copy of the GNU General Public License
+!     along with this program; if not, write to the Free Software
+!     Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+!
+!     This function is a copy of the function for calculation and printout of the lift and drag forces
```
MakisH, line 19: Does this file come directly from CalculiX, or does it have modifications?
Is this comment part of the original file, or of the preCICE-specific modifications?
If I wanted to upgrade the adapter to a new CalculiX version, what changes in this file should I port?

And, even if clear from the name: What is this function meant to do?

Same in the C3D8 file.

getc3d4elementgausspointcoords.f
```diff
@@ -0,0 +1,86 @@
...
+     &    co, kon, ipkon, elem_gp_coord)
+
+         implicit none
+
+         !Input/Output variables
+         integer     :: nelem            ! Number of elements
+         integer     :: elem_ids(*)      ! Element IDs, need to be adjusted for Fortran
+         real(8)     :: co(3,*)          ! Nodal coordinates of all nodes
+         integer     :: kon(*)
+         integer     :: ipkon(*)
```
MakisH, line 31: Description missing.

Same in the C3D8 file.

linstatic_precice.c
MakisH, line 1: A summary of the preCICE-related changes would be helpful when upgrading to newer CalculiX versions.

adapter/PreciceInterface.c
```diff
@@ -73,9 +74,9 @@ void Precice_AdjustSolverTimestep(SimulationData *sim)
     fflush(stdout);
 
     // For steady-state simulations, we will always compute the converged steady-state solution in one coupling step
-    *sim->theta  = 0;
-    *sim->tper   = 1;
-    *sim->dtheta = 1;
+    *sim->theta = 0;
+    *sim->tper  = 1;
+    //*sim->dtheta = 1;
```
MakisH, line 79: Why removed?
Maybe fully remove?

adapter/PreciceInterface.c
```diff
@@ -315,7 +380,7 @@ void Precice_WriteCouplingData(SimulationData *sim)
                                   sim->istartset,
                                   sim->iendset,
                                   sim->ipkon,
-                                  sim->lakon,
+                                  *sim->lakon,
```
MakisH, line 391: Is `lakon` a pointer?
Was this wrong before, or needed now because the type changed?
Does this need to be updated anywhere else?

adapter/PreciceInterface.c
```diff
@@ -804,6 +1000,7 @@ void PreciceInterface_FreeData(PreciceInterface *preciceInterface)
   free(preciceInterface->writeData);
   free(preciceInterface->elementIDs);
   free(preciceInterface->faceIDs);
+  free(preciceInterface->nodeIDs);
```
MakisH: How come we previously had `elementIDs` but not `nodeIDs`?

MakisH (commented):
A first round of comments after reading through most files and mostly skimming through `adapter/PreciceInterface.c` and `linstatic_precice.c`. Let's go through these files together as well.

Some of my comments are not necessarily something to act upon, just something for my own understanding.

Makefile
IshaanDesai: > Do I understand correctly that these files are completely written for the adapter, and not included in the original CalculiX source code?

~~Yes. These files are purely adapter-specific functionality. Moving them into the `adapter/` folder makes sense and I will do so.~~ On having a second look, the files have functionality which is directly copied from the CalculiX source code. So the files are intended for use in the adapter, but are not strictly code written by an adapter developer. The files are similar to `nonlingeo_precice.c` or even `ccx_2.20.c` for that matter.

My plan now is to remove the `linstatic_precice.c` entirely, because it is not serving the intended purpose of being a static solver for a multiscale scenario. I am trying to add the commands to read and write material data (stresses and strains) in `nonlingeo_precice.c` so that we do not have to maintain a second solver in the adapter.

adapter/CCXHelpers.c
IshaanDesai: I don't think we need a bound check for `k` here because we are iterating from the start to the end of an element set. An element set is like a tag for a set of finite elements which are part of the discretized mesh. At least right now I do not see a risk that `k` will overflow the number of elements in a set, which should be the size of `elements.`

adapter/CCXHelpers.c
IshaanDesai: Good catch with the data type! I think using the CalculiX-specific `ITG` would be a better choice.

adapter/CCXHelpers.h
IshaanDesai: `eei` is a pointer to the strain values inside CalculiX. I will change the description.

adapter/CCXHelpers.c
MakisH: Usual issue, but since we are anyway in C, there is not much we can do better, and I guess the codebase is already full of such potential out-of-bounds issues.

MinhazPX:
I could be able to build the adapter by making these changes to the `ConfigReader.cpp`

`if (config["participants"][participantName]["interfaces"][I]["elements-mesh”]) {
interface.elementsMeshName = strdup(config["participants"][participantName]["interfaces"][i]["elements-mesh"].get_value<std::string>().c_str());
}`

to 

`if (config["participants"][participantName]["interfaces"][i].contains("elements-mesh")) {
interface.elementsMeshName = strdup(config["participants"][participantName]["interfaces"][i]["elements-mesh"].get_value<std::string>().c_str());
}`

Hope this helps.

IshaanDesai:
Thanks @MinhazPX for the suggestion. The build works now. I will look into the `Segmentation fault` problem you are facing.

IshaanDesai:
While trying to get this to work together with @MinhazPX, I noticed that the data structure `stx` from `nonlingeo_precice.c` cannot be modified externally. Doing this corrupts the internal memory, and CalculiX is unable to handle the change when `results()` are called. One alternative is to modify the data `sti` instead. But this is problematic, as CalculiX internally uses `sti` to store the converged stress values. This means the only way to modify stresses from outside is to write a UMAT and provide it to `results()`.

It seems the current way of handling stress won't work.

adapter/PreciceInterface.c
MakisH: This is indeed a bug: https://github.com/precice/calculix-adapter/pull/173

Issue #136, opened 2025-03-21
Inconsistent use of `dim` and `dimCCX`


MakisH:
For example:

https://github.com/precice/calculix-adapter/blob/1690a6736bd90e9f963bf2eee1fc6eb7112d26cf/adapter/PreciceInterface.c#L376-L410

I guess there must be a consistency check earlier, but this is a bit confusing and potentially dangerous.

PR #137, opened 2025-03-21
Extend velocities writing for the 2D3D case

MakisH:
It looks like when reading velocities, we were not performing the same tests for the pseudo-2D case as in the rest, most probably as a result of how and when the writing of the various fields has been added to the codebase.

@IshaanDesai I think you implemented the 2D-3D mapping a while ago. Do you think this makes sense? Do you see any other places where we have similar issues?

Also related to #136: I changed one `dim` to `dimCCX`, for consistency at least in the same `case` branch.

Bug reported and fix proposed by a user in https://precice.discourse.group/t/error-happend-when-writing-velocites-from-calculix/2270

### DIFF
--- a/adapter/PreciceInterface.c
+++ b/adapter/PreciceInterface.c
@@ -394,8 +394,13 @@ void Precice_WriteCouplingData(SimulationData *sim)
         printf("Writing DISPLACEMENTDELTAS coupling data.\n");
         break;
       case VELOCITIES:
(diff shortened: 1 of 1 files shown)


IshaanDesai (approved):
This makes sense :+1: The quantities `POSITIONS` and `FORCES` do not have logic for 2D-3D cases. For `FORCES` we could use `conservativeVectorWrite`. For `POSITIONS` it is not clear what would be the correct option between consistent and conservative.

GongChen-Charlie:
Hi, @MakisH 
This is my GitHub account, I'm glad I could help in this 2D3D case.

PR #139, opened 2025-04-05
Makefile improvements: compiler paths and version-aware flags

abzrgzdh:
Hi,

This is a PR based on [this discussion](https://precice.discourse.group/t/problem-linking-adapted-calculix-against-yaml-cpp-macos/2264). The main goal is to update `:/Makefile` so that the adapter can be built on macOS as well.

In short, the adaptor couldn't link against `yaml-cpp`, because the compilation was done using `gcc`, `g++` and `-lstdc++`. Changing these to `clang`, `clang++`, and `-lc++` fixed that.

Other compilation issues encountered included:
- (C) Implicit declaration of functions
- (C) Incompatible pointer types
- (Fortran) -fallow-argument-mismatch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,9 @@ LIBS = \
 #CFLAGS = -g -Wall -std=c++11 -O0 -fopenmp $(INCLUDES) -DARCH="Linux" -DSPOOLES -DARPACK -DMATRIXSTORAGE
 #FFLAGS = -g -Wall -O0 -fopenmp $(INCLUDES)
 
-CFLAGS = -Wall -O3 -fopenmp $(INCLUDES) -DARCH="Linux" -DSPOOLES -DARPACK -DMATRIXSTORAGE -DUSE_MT -Wno-implicit-function-declaration -Wno-incompatible-pointer-types
+CFLAGS  = -Wall -O3 -fopenmp $(INCLUDES)
+CFLAGS += -DARCH="Linux" -DSPOOLES -DARPACK -DMATRIXSTORAGE -DUSE_MT
+CFLAGS += -Wno-implicit-function-declaration -Wno-incompatible-pointer-types
 
 # OS-specific options
 UNAME_S := $(shell uname -s)
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ CFLAGS += -Wno-implicit-function-declaration -Wno-incompatible-pointer-types
 # OS-specific options
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
-	CC = /usr/local/bin/gcc
+	CC = clang
 	LIBS += -lc++
 else
 	CC = mpicc
--- a/Makefile
+++ b/Makefile
@@ -52,9 +52,11 @@ CFLAGS += -Wno-implicit-function-declaration -Wno-incompatible-pointer-types
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
 	CC = clang
+	CXX = clang++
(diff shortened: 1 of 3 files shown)


IshaanDesai:
@abzrg apologies for not looking into this PR sooner. We recently removed yaml-cpp as a dependency. Can you please update your feature branch with respect to the develop branch? I will then review it.

abzrg:
Thank you @IshaanDesai for letting me know. I rebased my changes on the latest changes of `develop`.

Makefile
```diff
@@ -37,26 +37,34 @@ INCLUDES = \
...
 # OS-specific options
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
-	CC = /usr/local/bin/gcc
+	CC = clang
+	CXX = clang++
+	LIBS += -lc++
 else
 	CC = mpicc
+	CXX = g++
```
fsimonis: Lets keep it consistent.

```suggestion
	CXX = mpic++
```

Makefile
```diff
@@ -37,26 +37,34 @@ INCLUDES = \
...
+	LIBS += -lstdc++
 endif
 
 FFLAGS = -Wall -O3 -fopenmp $(INCLUDES) ${ADDITIONAL_FFLAGS} -Wno-implicit-function-declaration
-# Note for GCC 10 or newer: add -fallow-argument-mismatch in the above flags
+GCC_VERSION_MAJOR := $(shell $(CC) -dumpversion | cut -d. -f1)
+ifeq ($(shell [ $(GCC_VERSION_MAJOR) -ge 10 ] && echo yes),yes)
+	FFLAGS += -fallow-argument-mismatch
+endif
```
fsimonis: This only works for and is required for gfortran. So this check is incomplete.

Makefile
abzrgzdh: [83b6529](https://github.com/precice/calculix-adapter/pull/139/commits/83b652986b3368680e8a2293490dce17836bfdee)

Makefile
abzrgzdh: [b372d9a](https://github.com/precice/calculix-adapter/pull/139/commits/b372d9a2f75b494112977726e500faee2ce7041d)

Makefile
```diff
@@ -37,29 +37,40 @@ INCLUDES = \
...
 endif
 
-FFLAGS = -Wall -O3 -fopenmp $(INCLUDES) ${ADDITIONAL_FFLAGS} -Wno-implicit-function-declaration
-# Note for GCC 10 or newer: add -fallow-argument-mismatch in the above flags
 FC = mpifort
 # FC = mpif90
 # FC = gfortran
+FFLAGS = -Wall -O3 -fopenmp $(INCLUDES) ${ADDITIONAL_FFLAGS} -Wno-implicit-function-declaration
+FC_NAME := $(shell $(FC) --version 2>/dev/null | head -n 1 | cut -d' ' -f1)
+ifeq ($(FC_NAME),GNU)
```
fsimonis: Less error prone
```suggestion
ifeq ($(findstring GNU,$(shell $(FC) --version)),GNU)
```

abzrgzdh: I've just noticed I can just commit the suggestion using the button.

fsimonis (approved):

MakisH (approved):
Thank you for the update! This is a very useful contribution, independent of the original reason (yaml-cpp was removed in https://github.com/precice/calculix-adapter/pull/143).

I like the version-aware `-fallow-argument-mismatch`, which saves the whole hard-coding. We should also update the documentation and the respective CI scripts to remove that (I will do).

PR #140, opened 2025-04-11
CI: Remove ubuntu-20.04, add ubuntu-24.04

MakisH:
The `ubuntu-20.04` runners are scheduled for removal soon. At the same time, we are not testing the build on Ubuntu 24.04.

This PR removes 20.04 and adds 24.04 in the `ubuntu_build.yml` workflow.

### DIFF
--- a/.github/workflows/ubuntu_build.yml
+++ b/.github/workflows/ubuntu_build.yml
@@ -14,13 +14,13 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-22.04, ubuntu-24.04]
         include:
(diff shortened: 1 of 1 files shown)


MakisH:
This seems to work, but I have not tried the generated package for Ubuntu 24.04.
@IshaanDesai could you maybe try and give me a review here, since you are lately working with this adapter?

IshaanDesai (approved):
I tried the adapter build on Ubuntu 24.04 and it is working :+1:

MakisH:
Triggered the workflow, we now have an Ubuntu 24.04 package in the latest release: https://github.com/precice/calculix-adapter/releases/tag/v2.20.1

PR #143, opened 2025-07-04
Migrate to header-only yaml

### LINKED ISSUE #141: Use a header-only yaml library
A very common problem users face is that the YAML library cannot be found or loaded at runtime.

We could switch to a header-only library, include it in the adapter source and compile it into the adapter itself.

This should eliminate this problem.

fsimonis:
This PR replaces the yaml-cpp dependency with the MIT licensed header-only YAML library https://github.com/fktn-k/fkYAML.

Problems installing and finding yaml-cpp are common and the header-only variant will solve a range of troubleshooting steps.

This also reduced the diff of the Makefiles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,9 @@
 # Set the following variables before building:
 # Path to original CalculiX source (e.g. $(HOME)/ccx_2.xx/src )
 CCX_VERSION			= 2.20
-CCX             = $(HOME)/CalculiX/ccx_$(CCX_VERSION)/src
+CCX             = CalculiX/ccx_$(CCX_VERSION)/src
 
-### Change these if you built SPOOLES, ARPACK, or yaml-cpp from source ###
+### Change these if you built SPOOLES, or ARPACK from source ###
 # SPOOLES include flags (e.g. -I$(HOME)/SPOOLES.2.2 )
 SPOOLES_INCLUDE   = -I/usr/include/spooles/
 # SPOOLES library flags (e.g. $(HOME)/SPOOLES.2.2/spooles.a)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -3,9 +3,9 @@
 # Set the following variables before building:
 # Path to original CalculiX source (e.g. $(HOME)/ccx_2.xx/src )
 CCX_VERSION			= 2.20
-CCX             = $(HOME)/CalculiX/ccx_$(CCX_VERSION)/src
+CCX             = CalculiX/ccx_$(CCX_VERSION)/src
```
IshaanDesai: Why is `$(HOME)` removed? Wouldn't this break the path to find the CalculiX sources?

adapter/fkYAML.hpp
IshaanDesai, line 1: Would we have to update this file when a newer version of YAML is required?

IshaanDesai (commented):
Thanks for this. It is an important step in increasing the usability of the adapter. I just have a few queries.

Makefile
fsimonis: Good catch. Sorry this was a local change of mine.

```suggestion
CCX             = $(HOME)/CalculiX/ccx_$(CCX_VERSION)/src
```

adapter/fkYAML.hpp
fsimonis: Do you think that there will be a need for this? An update is a single download.

IshaanDesai: Shouldn't be necessary too often.

IshaanDesai (approved):
I tested this with the perpendicular-flap and everything seems to work :+1:

PR #146, opened 2025-09-29
Functionality for volumetric coupling on quadrature points of a mesh

IshaanDesai:
This PR adds preliminary functionality to extract integration points from a provided mesh and use them as coupling vertices. The functionality is extracted from https://github.com/precice/calculix-adapter/pull/135 because it is deemed to be sufficient as a standalone feature. Integration points of a mesh having C3D8 (Hex) or C3D4 (Tet) elements can be extracted.


IshaanDesai:
The tutorial cases to be added via https://github.com/precice/tutorials/pull/653 tests this functionality.

getc3d4elementgausspointcoords.f
MakisH, line 1: I assume that these two files contain changes compared to the original CalculiX files. A comment on the modifications would be very helpful when porting to later versions.

Was the functionality provided by the deleted file never needed, or is it just substituted by the two new files?

MakisH (commented):
Some first comments, I have not yet checked if it works with the given tutorial.

If #135 is needed for https://github.com/precice/tutorials/pull/653, how can https://github.com/precice/tutorials/pull/653 be a test case for this reduced version? How would the behavior change?

How easy would it be to implement another, simpler tutorial, such as https://precice.org/tutorials-volume-coupled-diffusion.html?

getc3d4elementgausspointcoords.f
IshaanDesai: > I assume that these two files contain changes compared to the original CalculiX files. A comment on the modifications would be very helpful when porting to later versions.

Yes, the contents of both the new Fortran files is essentially copied from the CalculiX source code. Additionally I put some of my own code in there. I will add comments in the file.

> Was the functionality provided by the deleted file never needed, or is it just substituted by the two new files?

The functionality in the deleted file was not used. I tried to use it, but the Gauss point coordinates returned by it were not correct. Hence the decision to remove the file and add two separate files for two different elements. it id worth noting that in the future, these two files can be merged.

IshaanDesai:
I think I did not frame my initial comment correctly. This PR contains a part of the functionality which together with the contents of https://github.com/precice/calculix-adapter/pull/135 can be used to run the two-scale notch tutorial case. Standalone this PR cannot be tested, because we do not have a volumetric coupling case with CalculiX.

We can either decide to merge this PR based on a code review, or think about adding a simple volumetric coupling tutorial case. I can look up the CalculiX tutorials to try to find a case which we can easily modify.

MakisH:
In principle, we could merge this without a test case, hoping that someone (else?) in the future might pick into that again. But I foresee a lot of "theoretically, it supports volume coupling, but we have no example / do not really know".

I would suggest to think a bit more of a simple tutorial case, ideally reusing one of the current ones, even if it does not have physical motivation.

IshaanDesai:
> In principle, we could merge this without a test case, hoping that someone (else?) in the future might pick into that again. But I foresee a lot of "theoretically, it supports volume coupling, but we have no example / do not really know".

Actually the follow-up PR to this would be #135 which uses the functionality. I simply extracted the functionality from that PR to keep features separate and have a better git history.

> I would suggest to think a bit more of a simple tutorial case, ideally reusing one of the current ones, even if it does not have physical motivation.

Let me think a bit more about this.

MakisH:
By the way, we should also extend the documentation of the adapter to mention the exact capabilities and limitations of this.

IshaanDesai:
> By the way, we should also extend the documentation of the adapter to mention the exact capabilities and limitations of this.

Yes. Right now the documentation is in https://github.com/precice/precice.github.io/pull/530. I can extract the parts relevant for this PR and prepare a new PR for the website.

MakisH:
> > By the way, we should also extend the documentation of the adapter to mention the exact capabilities and limitations of this.
> 
> Yes. Right now the documentation is in [precice/precice.github.io#530](https://github.com/precice/precice.github.io/pull/530). I can extract the parts relevant for this PR and prepare a new PR for the website.

It actually looks like that PR would be ready to merge if only it shifted a bit from the multiscale to the general volume coupling perspective, showing multiscale as one example.

IshaanDesai:
> It actually looks like that PR would be ready to merge if only it shifted a bit from the multiscale to the general volume coupling perspective, showing multiscale as one example.

I will change the language in that PR then, so we can proceed to merging it. I would also recommend merging this PR then.

docs/configure.md
```diff
@@ -45,6 +53,18 @@ On nodes-mesh:
...
+
+* stresses1to3 (components (1,1), (2,2), (3,3) of the stress tensor)
+* stresses4to6 (components (2,3), (1,3), (1,2) of the stress tensor)
+* cmat1 (components (1,1), (1,2), (1,3) of the material stiffness tensor)
+* cmat2 (components (1,4), (1,5), (1,6) of the material stiffness tensor)
+* cmat3 (components (2,2), (2,3), (2,4) of the material stiffness tensor)
+* cmat4 (components (2,5), (2,5), (3,3) of the material stiffness tensor)
+* cmat5 (components (3,4), (3,5), (3,6) of the material stiffness tensor)
+* cmat6 (components (4,4), (4,5), (4,6) of the material stiffness tensor)
+* cmat7 (components (5,5), (5,6), (6,6) of the material stiffness tensor)
```
MakisH: Open discussion in the original PR: https://github.com/precice/precice.github.io/pull/530#discussion_r2194504455

Current state: some cleanup is needed to focus on the feature of this PR vs the feature introduced with https://github.com/precice/calculix-adapter/pull/135.

Let's continue the discussion here.

getelementgausspointcoords.f
```diff
@@ -0,0 +1,113 @@
...
+!     This program is distributed in the hope that it will be useful,
+!     but WITHOUT ANY WARRANTY; without even the implied warranty of 
+!     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
+!     GNU General Public License for more details.
+!
+!     You should have received a copy of the GNU General Public License
+!     along with this program; if not, write to the Free Software
+!     Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+!
+!     This function is a copy of the function for calculation and printout of the lift and drag forces
```
MakisH, line 19: @IshaanDesai What function does this refer to? A link to the source code of CalculiX would make it easier to port this function to a newer version of CalculiX.

MakisH: Since I have a few other PRs depending on this one, let's add the comment as soon as possible after merging.

MakisH (approved):

getelementgausspointcoords.f
IshaanDesai: This function is kind of a patchwork of loops from the CalculiX source code. Previously, it was more of a direct copy, but I later removed unnecessary parts to shorten the file. I think updating this function for newer versions of CalculiX should be fairly straightforward.

PR #147, opened 2025-10-08
Add a CHANGELOG file

IshaanDesai:
Title is self-explanatory. I only added entries to the changelog **after** the last release.


MakisH:
Good point, thank you!

In other repositories, we use a system in which we collect changelog entries in a specific directory. This has the advantage of not running into merge conflicts (which will always happen).

I would rather introduce that system here first, and create a changelog file with the next release. There, we can also use https://keepachangelog.com/en/1.1.0/, and I could collect changelog entries from previous releases as well.

IshaanDesai:
> I would rather introduce that system here first, and create a changelog file with the next release. There, we can also use https://keepachangelog.com/en/1.1.0/, and I could collect changelog entries from previous releases as well.

Good point. I will add separate files in a dedicated folder.

IshaanDesai:
@MakisH I now put all the changelog entries since the last release in separate files, in `docs/changelog/`.

MakisH:
I would still like to do some work on the same branch, give me a few days.

IshaanDesai:
> I would still like to do some work on the same branch, give me a few days.

Shall we merge this branch?

MakisH (approved):
Taking another look, I think that this repository is not active enough to justify a `changelog-entries/` workflow, which would be a bit more annoying (one has to first get the number of the PR to create the file). We can always reintroduce such a workflow without any changes to the file.

I added a `CHANGELOG.md` based on the keeachangelog format, and transferred the past entries from the Releases here, adjusting for consistency. I also added numbers and links to all PRs (besides the one for v2.20, which I think the commits were scattered).

PR #156, opened 2026-07-20
Update instructions for modifying end time in .inp file

gdenayer:
Clarified instructions for modifying end time in the .inp file and added details about the time step size and STEP duration.


MakisH (approved):
Thank you for the documentation improvements; keep them coming! 🤗

